Issue #71, opened 2020-11-23
Port and Flow stats should returned as PortResult and FlowResult object instead of dictionary list


ashutshkumr:
According to API spec, one should be able to perform following:
```python
flow_results = api.get_flow_results(result.FlowRequest())
print(flow_results[0].frames_tx)
```

But it doesn't work and upon dumping, it's seen that it's a list of dictionary: 
```sh
[{'name': 'raw tx->rx', 'transmit': 'started', 'port_tx': 'tx', 'port_rx': 'rx', 'frames_tx': 1000, 'frames_rx': 1000, 'frames_tx_rate': 0.0, 'frames_rx_rate': 0.0, 'bytes_tx_rate': 0.0, 'bytes_rx_rate': 0.0, 'loss': 0.0}]
```

PR #91, opened 2020-11-26
Fixed https://github.com/open-traffic-generator/ixnetwork/issues/71

ashutshkumr:
We're now returning list of result.Port and result.Flow instead of flow dict and port dict.


ashutshkumr:
@ajbalogh I had a discussion with @alakendu and looks like this is something we might need to tell customers about. (Move from stat dictionary to object).

Hence, we'll need more discussion before including this one in `master`. I already made some changes above this one in `stabilize` branch, so merging it temporarily and will revert it if we decide to not proceed with it.

Issue #93, opened 2020-11-26
Add test for UDP header with fixed fields


ANISH-GOTTAPU:
- Add new test script inside tests/udp/
- Provide non-default, fixed values for all UDP fields.
- Validate that all the fields have attained expected values from captured frames.

PR #94, opened 2020-11-26
adding udp tests

ANISH-GOTTAPU:
Adding udp related test with ref #93

### DIFF
--- a/tests/udp/test_raw_traffic_with_fixed_udp_ports.py
+++ b/tests/udp/test_raw_traffic_with_fixed_udp_ports.py
@@ -1,0 +1,78 @@
+from abstract_open_traffic_generator.flow import *
+import pytest
+
+from abstract_open_traffic_generator import flow
+import utils
+
+
+@pytest.mark.parametrize('packets', [1000])
+@pytest.mark.parametrize('size', [74, 128, 1518])
+def test_raw_traffic_with_fixed_udp_ports(api, b2b_raw_config, size, packets):
+    """
+    Configure a raw udp flow with,
+    - fixed src and dst Port address
+    - 1000 frames of 74B,128B,1518B size each
+    - 10% line rate
+
+    Validate,
+    - tx/rx frame count is as expected
+    - all captured frames have expected src and dst Port address
+    """
+    f = b2b_raw_config.flows[0]
+
+    f.packet = [
+        flow.Header(
+            flow.Ethernet(
+                src=flow.Pattern('00:0c:29:1d:10:67'),
+                dst=flow.Pattern('00:0c:29:1d:10:71')
+                )
+            ),
+        flow.Header(
+            flow.Ipv4(
+                src=flow.Pattern("10.10.10.1"),
+                dst=flow.Pattern("10.10.10.2")
+                )
+            ),
+        flow.Header(
+            flow.Udp(
+                src_port=flow.Pattern("6"),
+                dst_port=flow.Pattern("7")
+            )
+        ),
+    ]
+    f.duration = flow.Duration(flow.FixedPackets(packets=packets))
+    f.size = flow.Size(size)
+    f.rate = flow.Rate(value=10, unit='line')
+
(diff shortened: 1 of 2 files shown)


tests/udp/test_raw_traffic_with_fixed_udp_ports.py
```diff
@@ -0,0 +1,78 @@
+from abstract_open_traffic_generator.flow import *
```
ashutshkumr, line 1: Why do we need this ?

tests/udp/test_raw_traffic_with_fixed_udp_ports.py
```diff
@@ -0,0 +1,78 @@
...
+import pytest
+
+from abstract_open_traffic_generator import flow
+import utils
+
+
+@pytest.mark.parametrize('packets', [1000])
```
ashutshkumr, line 8: Let's get rid of packets and size fixtures as they increase test runtime a lot. I did the same for ethernet/* scripts.
We'll have separate, dedicated tests to cover these.

tests/udp/test_raw_traffic_with_fixed_udp_ports.py
```diff
@@ -0,0 +1,78 @@
...
+@pytest.mark.parametrize('size', [74, 128, 1518])
+def test_raw_traffic_with_fixed_udp_ports(api, b2b_raw_config, size, packets):
```
ashutshkumr, line 10: We can get rid of `raw_traffic` in test and file names.

tests/udp/test_raw_traffic_with_fixed_udp_ports.py
```diff
@@ -0,0 +1,78 @@
...
+            ),
+        flow.Header(
+            flow.Ipv4(
+                src=flow.Pattern("10.10.10.1"),
+                dst=flow.Pattern("10.10.10.2")
+                )
+            ),
+        flow.Header(
+            flow.Udp(
+                src_port=flow.Pattern("6"),
```
ashutshkumr, line 38: This is not really a technical issue, but let's use ports outside 1-1024 range (something that will require more than 1 byte).

tests/udp/test_raw_traffic_with_fixed_udp_ports.py
```diff
@@ -0,0 +1,78 @@
...
+
+    captures_ok(api, b2b_raw_config, size)
+
+
+def stats_ok(api, packets):
+    """
+    Returns true if stats are as expected, false otherwise.
+    """
+    port_results, flow_results = utils.get_all_stats(api)
+
```
ashutshkumr, line 60: Check total frame size as well.

ashutshkumr (changes requested):

tests/udp/test_raw_traffic_with_fixed_udp_ports.py
ANISH-GOTTAPU: was added when testing, removed now

tests/udp/test_raw_traffic_with_fixed_udp_ports.py
ANISH-GOTTAPU: Have removed multiple iterations but has kept a single value instead of hardcoding inside the test

tests/udp/test_raw_traffic_with_fixed_udp_ports.py
ANISH-GOTTAPU: done

tests/udp/test_raw_traffic_with_fixed_udp_ports.py
ANISH-GOTTAPU: updated

tests/udp/test_raw_traffic_with_fixed_udp_ports.py
ANISH-GOTTAPU: updated

Issue #75, opened 2020-11-24
Retreiving correct pcap requires sending capture request twice


ashutshkumr:
Sending a capture request automatically stops the capture and returns a binary file in .pcap format. But in the first try, it always returns a .pcap file with 0 packets.

Upon second try (without adding any wait) correct .pcap file is returned (with intended packet count).

alakendu:
I am not getting this issue when run this test script ..\ixnetwork\tests\test_capture_pcap.py in my system.
Could you please specify test script, chassis card and port and python version where you able to reproduced it.

ashutshkumr:
In the same script, if you increase the packet count from 100 to 1000, the issue will reproduce.

alakendu:
Internally setting max time_out to 90sec with 3sec polling interval. But this limit will may overshoot for some corner case. Please think if we can expose time_out within CaptureRequest.

Issue #73, opened 2020-11-23
Setting intended speed as part of layer1 config is not reflected on target ports


ashutshkumr:
To reproduce this against VMOne ports, run following from stabilize branch: https://github.com/open-traffic-generator/ixnetwork/blob/stabilize/tests/ethernet/test_fixed_mac_addrs.py

In order to change speed, modify `speed` option in tests/settings.json (from 1G to 100MB FD). The modified speed is not reflected after running the tests.

alakendu:
This issue we are addressing VM port speed. We will prioritize according to the customer require ports and speed.

Issue #72, opened 2020-11-23
Some attributes in port and flow results are arbitarily missing


ashutshkumr:
Upon running unidirectional b2b test with one flow, it's seen that not all `columns` are returned for port and flow stats.

```sh
# port stats - tx
{'name': 'tx', 'location': '10.74.44.174;1;1;connected', 'link': 'up', 'capture': 'stopped', 'frames_tx': 0, 'frames_rx': 0, 'frames_tx_rate': 0, 'frames_rx_rate': 0, 'bytes_tx': 0, 'bytes_rx': 0, 'bytes_tx_rate': 0, 'bytes_rx_rate': 0}

# port stats - rx - most stats missing compared to tx
{'name': 'rx', 'location': '10.74.44.174;1;2;connected', 'link': 'up', 'capture': 'stopped'}

# flow stats - bytes_rx count is missing
{'name': 'raw tx->rx', 'transmit': 'started', 'port_tx': 'tx', 'port_rx': 'rx', 'frames_tx': 1000, 'frames_rx': 1000, 'frames_tx_rate': 0.0, 'frames_rx_rate': 0.0, 'bytes_tx_rate': 0.0, 'bytes_rx_rate': 0.0, 'loss': 0.0}
```

Another note here: `location` returns the location concatenated with string `connected` even though the original location string shouldn't have been modified.

alakendu:
Please attached/ specify location of that script

ashutshkumr:
This script fetches port and flow stats. You'll see the issue upon dumping them: https://github.com/open-traffic-generator/tests/blob/master/samples/test_raw_eth.py (steps on how to run are in readme.md)

BTW, most of the new issues are now being tracked as part of this project board and the assignee will be decided upon further discussion: https://github.com/open-traffic-generator/ixnetwork/projects/2

ashutshkumr:
There were two issues:
- For flow stats, bytes_rx was not being fetched (instead byte rate stats were being fetched)
- For port stats, exception was raised everytime PFC stats were being extracted for first port, even when those weren't visible in stat view. It never got a chance to proceed to next ports.

Fixed in https://github.com/open-traffic-generator/ixnetwork/commit/d2de08ec36cffd1a229a17dc3d1379d0a754127f

PR #110, opened 2020-12-01
Include changes from stabilize

ashutshkumr:
## Summary of changes:
- Introduced `tests/utils` to refactor planned tests (be able to validate stats and capture for each relevant test)
- Added planned tests - WIP
- Added GTPv1 header - WIP
- Fixed issue with missing/broken stat attributes - https://github.com/open-traffic-generator/ixnetwork/issues/72
- Fixed issue with incorrect pcap received intermittently - https://github.com/open-traffic-generator/ixnetwork/issues/75
- Fixed issue with l1 speed not being set for VM ports- https://github.com/open-traffic-generator/ixnetwork/issues/73
- Added requirements.txt to record pre-build dependencies
- Added steps for dev setup in contributing.md
- Modify publish.yml to prepare setup for running tests per changes above
- Expose `tests/settings.json` to allow directly configuring it (and other config files specific to linting and pytest)
- Modify README.md to include working code and correct formatting

## Things to note during merge:
- Upgrade version and run tests against newly published package (to check for any regression)


ajbalogh (approved):
All ci/cd steps have passed.
I don't agree with not using pandas.
We should be steering customers to best practices and an independent utils/common files is not the best approach.
We should be providing an approach that is based on pytest of which fixtures are central feature versus a custom utility library.
If the models are changed we should align with the models (ie port stats, flow stats).

winstonliu1111 (approved):
LGTM

ashutshkumr:
> All ci/cd steps have passed.
> I don't agree with not using pandas.
> We should be steering customers to best practices and an independent utils/common files is not the best approach.
> We should be providing an approach that is based on pytest of which fixtures are central feature versus a custom utility library.
> If the models are changed we should align with the models (ie port stats, flow stats).

I agree with moving most (relevant) utils items to fixtures, and will track it as seperate issue (if that's ok).
We can't completely retire utils yet though because currently it is helping us with rapid migrations without needing to batch update test scripts.

Regarding `pandas`, I have nothing against it and it's a great library for working with different datasets.
I just removed it from readme sample because equivalent code using `in-built` seemed more concise and intuitive. i.e.

```python
request = result.PortRequest(column_names=['name', 'location', 'frames_tx', 'frames_rx'])
results = api.get_port_results(request)

# pandas
df = pandas.DataFrame.from_dict(results)
df.frames_tx.sum()

# in-built
sum(r['frames_tx'] for r in results)
```

ankur-sheth (approved):

winstonliu1111:
regarding andy/ashutosh's comment on pandas, I think pandas is super useful for slice/dice and when you deal with a table.  the in-built mechanism sort of breaks down and efficiency when you have to access more than 1 metric, or perform different operations over the same metric.  maybe show both ways?  and show pandas where that really shines.

Issue #98, opened 2020-11-30
Add test for UDP header with Counter Pattern for UDP src,dst,length and checksum


ANISH-GOTTAPU:
- Add new test script inside tests/udp/
- For each pattern type in list
       - Provide non-default counter pattern values for src port, dst port, length and checksum
       - Validate that src port, dst port, length and checksum are as expected from captured frame

ashutshkumr:
@ANISH-GOTTAPU we won't need to do it for length and checksum. Please open another issue for those (check IPv4 test plan for length, for reference)

ANISH-GOTTAPU:
@ashutshkumr : As discussed, I will add different issue to test length that is being set as per frame size. This test would check the user passed length and checksum values

ANISH-GOTTAPU:
Closed with #113

Issue #95, opened 2020-11-27
Add test for UDP header with List Pattern for UDP src,dst,length and checksum


ANISH-GOTTAPU:
- Add new test script inside tests/udp/
- For each pattern type in list
       - Provide non-default list values for src port, dst port, length and checksum
       - Validate that src port, dst port, length and checksum are as expected from captured frame

ANISH-GOTTAPU:
Closed with #113

PR #112, opened 2020-12-02
Add udp tests

ANISH-GOTTAPU:
Tests for #95 & #98


ANISH-GOTTAPU:
will be raising PR against master branch

PR #113, opened 2020-12-02
add udp tests

ANISH-GOTTAPU:
adding udp tests for #95  & #98

### DIFF
--- a/tests/udp/test_udp_header_with_counter.py
+++ b/tests/udp/test_udp_header_with_counter.py
@@ -1,0 +1,72 @@
+import pytest
+import utils
+from abstract_open_traffic_generator import flow as Flow
+
+
+@pytest.mark.parametrize('packets', [100])
+@pytest.mark.parametrize('size', [74])
+def test_udp_header_with_counter(api, b2b_raw_config, size, packets):
+    """
+    Configure a raw udp flow with,
+    - Non-default Counter Pattern values of src and
+      dst Port address, length, checksum
+    - 100 frames of 74B size each
+    - 10% line rate
+
+    Validate,
+    - tx/rx frame count is as expected
+    - all captured frames have expected src and dst Port address
+    """
+    flow = b2b_raw_config.flows[0]
+
+    src_port = ('5000', '2', '10')
+    dst_port = ('6000', '2', '10')
+    length = ('35', '1', '2')
+    checksum = ('6', '1', '2')
+    flow.packet = [
+        Flow.Header(
+            Flow.Ethernet(
+                src=Flow.Pattern('00:0c:29:1d:10:67'),
+                dst=Flow.Pattern('00:0c:29:1d:10:71')
+            )
+        ),
+        Flow.Header(
+            Flow.Ipv4(
+                src=Flow.Pattern('10.10.10.1'),
+                dst=Flow.Pattern('10.10.10.2')
+            )
+        ),
+        Flow.Header(
+            Flow.Udp(
+                src_port=Flow.Pattern(
+                    Flow.Counter(start=src_port[0], step=src_port[1],
+                                 count=int(src_port[2]))
+                ),
+                dst_port=Flow.Pattern(
+                    Flow.Counter(start=dst_port[0], step=dst_port[1],
+                                 count=int(dst_port[2]), up=False)
+                ),
+                length=Flow.Pattern(
+                    Flow.Counter(start=length[0], step=length[1],
+                                 count=int(length[2]))
+                ),
+                checksum=Flow.Pattern(
+                    Flow.Counter(start=checksum[0], step=checksum[1],
+                                 count=int(checksum[2]))
+                )
+            )
+        ),
+    ]
+    flow.duration = Flow.Duration(Flow.FixedPackets(packets=packets))
+    flow.size = Flow.Size(size)
+    flow.rate = Flow.Rate(value=10, unit='line')
+
+    utils.apply_config(api, b2b_raw_config)
+    attrs = {
+        'UDP-Source-Port': src_port,
+        'UDP-Dest-Port': dst_port,
+        'UDP-Length': length,
+        'UDP-Checksum': checksum
+    }
+    utils.validate_config(api, 'udp', **attrs)
+
--- a/tests/udp/test_udp_header_with_fixed.py
+++ b/tests/udp/test_udp_header_with_fixed.py
@@ -1,0 +1,62 @@
+import pytest
+import utils
+from abstract_open_traffic_generator import flow as Flow
+
+
+@pytest.mark.parametrize('packets', [1000])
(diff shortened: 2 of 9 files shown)


ashutshkumr (changes requested):
Please check https://github.com/open-traffic-generator/ixnetwork/pull/115 once since some of the comment there applies here as well.

@Rangababu-R I find `@pytest.mark.parametrize('packets', [100])` instead of `packets = 100` much more useful (since we have the ability to add more combinations as we need). What do you think ?

ashutshkumr:
@ajbalogh, could you please review if the approach used in following file should be sufficient for unit tests:
- [Test script](https://github.com/open-traffic-generator/ixnetwork/pull/113/files#diff-ba11e35ddfdd8af74c46b32762197e6ec920d918114557252147dc155398ac92)
- [Restpy helpers](https://github.com/open-traffic-generator/ixnetwork/pull/113/files#diff-32194464ea6026f41781b732e67e5c3aa8990c3191027d05846455d25fc0b7a5)

This reduces overall per-test duration from almost 2m to less than 10s (thanks to @ANISH-GOTTAPU and @Rangababu-R ).

@winstonliu1111 @ankur-sheth, since above changes will introduce restpy-specific code for most test scripts and won't really be applicable to other implementations, do you suggest moving the existing e2e tests out to a separate repo ? Since,
- It's desirable to be able to run same suite against all implementations (and thus be free from impl-specific internals)
- We want this to be a reference suite for first-time users
- We can't run them per-push because the suite is increasingly causing higher build duration, but we need to come up with a way to run them separately, per-version-bump (probably using some kind of webhook).
- There's value in running them because that's where we found the issues in stats and capture (this and last week)

ajbalogh (approved):
Not a big fan of a utils file when a fixture will do.
    utils.wait_for(
        lambda: results_ok(api, size, packets), 'stats to be as expected'
    )

Why not a fixture that returns a function, no need for import utils, a user having to remember to provide a lambda.


This is what I use my tests:

@pytest.fixture()
def wait_for_transmit_stopped(api):
    def _wait_for_transmit_stopped():
        import abstract_open_traffic_generator.result as result
        import time
        while True:
            results = api.get_flow_results(result.FlowRequest())
            data_frame = pandas.DataFrame.from_dict(results)
            print(data_frame)
            if data_frame.transmit.str.match('^stopped$').sum() == len(
                    data_frame):
                break
            time.sleep(1)
        return data_frame

    return _wait_for_transmit_stopped

def test_flow_results(serializer, api, b2b_ipv4_flow_config, config_state,
                      flow_transmit_state, wait_for_transmit_stopped):
    """Demonstrates the following:
    - Pausing the script until traffic has stopped
    """
    config_state.config=b2b_ipv4_flow_config
    api.set_state(control.State(config_state))
    api.set_state(control.State(flow_transmit_state))

    data_frame = wait_for_transmit_stopped()

winstonliu1111:
@ashutshkumr saw your comment to Ankur and I.  i was looking at the links you sent where we're testing say some pkt is properly mapped to ixnetwork properly and relies on restpy, this seems more like a unit test that should stay in this repo. since it exercises the mapping of data model to ixnetwork models only?  the e2e tests that is common across implementations should probably go further and transmit/rx and not deal with the unit test assertions, and assert based on what the users care about which are either captured content or stats/metrics?

ajbalogh:
I agree with @ashutshkumr that there should be a separate repo for tests that are agnostic, demonstrate best practices. I also think they should include real world test cases.

The tests repo within ixnetwork should use restpy where possible, should be as fast as possible and try to provide full coverage especially across problem areas.

ashutshkumr:
@winstonliu1111 yes, correct. But I think we won't be making that change immediately. For now, as I had mentioned in the last meeting, we'll mark some e2e tests to be skipped and just focus on increasing unit test coverage next couple of days. But it eventually needs to be taken care of sooner.


@ajbalogh Thank you for examples. I have opened a new issue to track this: https://github.com/open-traffic-generator/ixnetwork/issues/118.

I must admit that fixtures only make sense to me for passing objects (functions or not) that have a certain lifecycle, most of which requires setup and teardown (because behind the scenes pytest takes care of the lifecycle for us, while providing more controls to manage it).

Beyond that, I still don't understand what's the value in writing factories for functions that do not have a lifecycle. It eventually ends up littering conftest and test arguments.

I mean, sure, if this is a best practice, it's an easy change. But why not just pass `utils` (or smaller sub packages) as a fixture ?

ankur-sheth:
@ashutshkumr - what is holding us up in moving the implementation agnostic tests to a separate repository?  Finalization of snappi?  We should start thinking about it as we're going to "finalize" snappi pretty quickly.

ajbalogh:
@ashutshkumr i agree with you, lets use the facilities of the package ie pytest provides fixture infrastructure so lets use it - the rest becomes choice - mine was just a sample, whatever you decide you should also add that to the contributing guide.

ashutshkumr:
@ankur-sheth there's nothing blocking per se. It's just that I think it'll make more sense to prioritize adding more coverage in following areas, before moving to more infra-centric work (which relocating and refactoring e2e test repo will require):
- Protocols headers
- Tx Scheduler
- PFC

We had planned to cover most commonly used scenarios for all above, by 4th Dec, Fri.

tests/udp_validate_with_restpy/test_udp_header_with_counter_restpy.py
```diff
@@ -0,0 +1,77 @@
...
+                    Flow.Counter(start=checksum[0], step=checksum[1],
+                                 count=int(checksum[2]))
+                )
+            )
+        ),
+    ]
+    flow.duration = Flow.Duration(Flow.FixedPackets(packets=packets))
+    flow.size = Flow.Size(size)
+    flow.rate = Flow.Rate(value=10, unit='line')
+
```
Rangababu-R, line 63: does these really required?

ashutshkumr: Rate is not required but size and duration is.

Num of packets affects field pattern expansion. Size may sometimes be useful to check whether protocol headers result in size auto-adjust.

Rangababu-R: @ashutshkumr Now the implementation is fetching the packet configuration from the packet fields instead of packet preview.
So, duration and size doesn't take any affect.

ashutshkumr: @Rangababu-R you're right, since we're only validating against config and not the preview, they're not required.

But as an aside, as long as it doesn't affect the test run duration, I'm fine with keeping them. It helps with reusablitly (copy-paste without adding/removing). What do you think ?

ashutshkumr (approved):

Issue #104, opened 2020-11-30
Generic API to return list of values for Counter up/down wards step


Rangababu-R:
- Create generic API for ipv4 Counter
- Create generic API for mac address Counter
- Create generic API for integer Counter

Rangababu-R:
Added generic API.

def mac_or_ip_addr_from_counter_pattern(start_addr, step, count, up, mac=True):
    """
    Example:
    mac_or_ip_addr_from_counter_pattern('10.1.1.1', '0.0.1.1', 2, True, False)
    returns: ['00:0C:29:E3:53:EA', '00:0C:29:E3:54:EA']
    mac_or_ip_addr_from_counter_pattern('10.1.1.1', '0.0.1.1', 2, True, False)
    teturns: ['10.1.1.1', '10.1.2.2']
    """

Issue #87, opened 2020-11-25
Add test for IPv4 header with varying src and dst IP


ashutshkumr:
- Add new test script inside tests/ipv4/
- For each pattern type in list, counter and random,
  - Provide non-default values for src and dst IP
  - Validate that src and dst IPs are as expected from captured frame

Rangababu-R:
Merged the code

PR #115, opened 2020-12-02
Ethernet and IP addresses pattern tests

Rangababu-R:
Added Test Cases for Ethernet addresses
and IP address pattern validation.
covers: Issues #104  and #87


ashutshkumr:
@Rangababu-R can you please check if python2 failur is legimitate on your setup.

tests/ethernet/eth.py
```diff
@@ -0,0 +1,23 @@
+import utils
```
ashutshkumr, line 1: Please move all common functions to utils itself. If you think it belongs to a specific separate module, feel free to create eth.py inside `utils/`

tests/ethernet/eth.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+def captures_ok(api, cfg, size, src, dst):
```
ashutshkumr, line 4: after moving it to utils, see if we can call it like so:
```python
utils.eth.captures_ok()
```

It's not impossible, but if it's too much effort for now, consider renaming it to add prefix.

tests/ethernet/eth.py
```diff
@@ -0,0 +1,23 @@
...
+    Returns normally if patterns in captured packets are as expected.
+    """
+    cap_dict = utils.get_all_captures(api, cfg)
+    src_list = []
+    dst_list = []
+    length = []
+    for k in cap_dict:
+        for b in cap_dict[k]:
+            dst_list.append(
+                ":".join(['{:02X}'.format(byt) for byt in b[0:6]])
```
ashutshkumr, line 15: wouldn't it make sense to move it to `mac_addr_from_bytes(b[0:6])` ? If we haven something similar for IPv4 as well, we should consider the same for it as well.

tests/ethernet/test_counter_mac_addr.py
```diff
@@ -0,0 +1,58 @@
+from abstract_open_traffic_generator import flow
+from abstract_open_traffic_generator.flow import Counter
```
ashutshkumr, line 2: let's be consistent here by simply using `flow.Counter` like all other flow components.

tests/ipv4/ipv4.py
```diff
@@ -0,0 +1,18 @@
+import utils
```
ashutshkumr, line 1: Requesting approach similar to discussed in mac addr section.

ashutshkumr (changes requested):
As I didn't get a chance to review yesterday the `generate*` functions. Here's what I suggest:
- All function name should follow consistent naming based on what they return and what pattern the accept. Something like `get_mac_addrs_from_counter_pattern()` or `mac_addrs_from_counter_pattern()`. I see some names which mixed counter with list (not intentional though).
- Reduce sample size of examples for `generate*` functions from 10 to 5 or 3. It's sufficient to showcase intent.
- Add functions `mac_addr_to_num()` and `num_to_mac_addr()`. Similarly for IPv4 addr. The use of `ipaddr` library isn't quite justified IMHO. The `generate` functions should be using those functions, resulting in a more reasonable code-block.
- We refactored `stats_ok` out of tests. Why not just refactor `wait_for` and call it `wait_for_stats_ok` ?

tests/ethernet/eth.py
Rangababu-R: As this was added at protocol level, did not feel to add as definition.
will move to definition

Rangababu-R:
> 
> 
> As I didn't get a chance to review yesterday the `generate*` functions. Here's what I suggest:
> 
>     * All function name should follow consistent naming based on what they return and what pattern the accept. Something like `get_mac_addrs_from_counter_pattern()` or `mac_addrs_from_counter_pattern()`. I see some names which mixed counter with list (not intentional though).
> 
>     * Reduce sample size of examples for `generate*` functions from 10 to 5 or 3. It's sufficient to showcase intent.
> 
>     * Add functions `mac_addr_to_num()` and `num_to_mac_addr()`. Similarly for IPv4 addr. The use of `ipaddr` library isn't quite justified IMHO. The `generate` functions should be using those functions, resulting in a more reasonable code-block.
> 
>     * We refactored `stats_ok` out of tests. Why not just refactor `wait_for` and call it `wait_for_stats_ok` ?

Done changes with restpy and removed ipaddr lib and made counter pattern local

Issue #126, opened 2020-12-08
Add markers to skip most e2e tests (since they should be moved out of repo)


ashutshkumr:

ashutshkumr:
https://github.com/open-traffic-generator/ixnetwork/pull/132

Issue #118, opened 2020-12-03
Move relevant items from utils to fixtures


ashutshkumr:
Comments from Andy:

Not a big fan of a utils file when a fixture will do.
```
    utils.wait_for(
        lambda: results_ok(api, size, packets), 'stats to be as expected'
    )
```
Why not a fixture that returns a function, no need for import utils, a user having to remember to provide a lambda.

This is what I use my tests:
```
@pytest.fixture()
def wait_for_transmit_stopped(api):
    def _wait_for_transmit_stopped():
        import abstract_open_traffic_generator.result as result
        import time
        while True:
            results = api.get_flow_results(result.FlowRequest())
            data_frame = pandas.DataFrame.from_dict(results)
            print(data_frame)
            if data_frame.transmit.str.match('^stopped$').sum() == len(
                    data_frame):
                break
            time.sleep(1)
        return data_frame

    return _wait_for_transmit_stopped

def test_flow_results(serializer, api, b2b_ipv4_flow_config, config_state,
                      flow_transmit_state, wait_for_transmit_stopped):
    """Demonstrates the following:
    - Pausing the script until traffic has stopped
    """
    config_state.config=b2b_ipv4_flow_config
    api.set_state(control.State(config_state))
    api.set_state(control.State(flow_transmit_state))

    data_frame = wait_for_transmit_stopped()

```

ashutshkumr:
https://github.com/open-traffic-generator/ixnetwork/pull/132

PR #132, opened 2020-12-08
Utils refactor

ashutshkumr:
Fixes:
- https://github.com/open-traffic-generator/ixnetwork/issues/126
- https://github.com/open-traffic-generator/ixnetwork/issues/118

### DIFF
--- a/tests/udp/test_udp_header_with_fixed_length_checksum_e2e.py
+++ b/tests/udp/test_udp_header_with_fixed_length_checksum_e2e.py
@@ -3,13 +3,8 @@ import utils
 from abstract_open_traffic_generator import flow as Flow
 
 
-@pytest.mark.skip(reason="skip until moved to other repo")
-@pytest.mark.parametrize('packets', [1000])
-@pytest.mark.parametrize('size', [74])
-def test_udp_header_with_fixed_length_checksum_e2e(api,
-                                                     b2b_raw_config,
-                                                     size,
-                                                     packets):
+@pytest.mark.e2e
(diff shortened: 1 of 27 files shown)


tests/udp/test_udp_header_with_fixed_length_checksum_e2e.py
```diff
@@ -3,13 +3,8 @@
 from abstract_open_traffic_generator import flow as Flow
 
 
-@pytest.mark.skip(reason="skip until moved to other repo")
-@pytest.mark.parametrize('packets', [1000])
```
ANISH-GOTTAPU, line 7: I prefer the packets and size to be parametrize for e2e tests, would be easy to increase test coverage if more unit test is required. But if tests to be look consistent, this is also fine for me

ANISH-GOTTAPU (approved):

Rangababu-R (approved):
The only change by making the utils as fixture, will avoid the import statement and pass as argument.
looks good to me

tests/udp/test_udp_header_with_fixed_length_checksum_e2e.py
ashutshkumr: I agree with parametrizing those variables. The problem is, should we or should we not parametrize other variables in the func as well. We need to converge on following:
- choose general variables that should be parameterized for all tests
- since those variables are pretty common across all tests, is there a way to control the parametrization from somewhere central rather than specifying them in the test script ? (pytest has something similar I guess)

Point is, we want to do it, but probably there's a better way than embedding same snippet in each test script.
Please raise an issue to specifically track this.

PR #143, opened 2020-12-09
vlan and ipv4 tests

Rangababu-R:
vlan test cases and ipv4 fields testcases
with restpy validation

### DIFF
--- a/tests/ipv4/test_counter_ipv4_fields.py
+++ b/tests/ipv4/test_counter_ipv4_fields.py
@@ -1,0 +1,155 @@
+from abstract_open_traffic_generator import flow
+
+
+def test_counter_ip_fields(api, b2b_raw_config, utils):
+    """
+    Configure a raw IPv4 flow with,
+    - counter pattern for the fields
+      header len, total len, identification,
+      reserved, don't fragment, more fragment,
+      fragment offset, time to live, protocol,
+      header checksum
+
+    Validate,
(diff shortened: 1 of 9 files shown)


tests/ipv4/test_counter_ipv4_fields.py
```diff
@@ -0,0 +1,155 @@
+from abstract_open_traffic_generator import flow
+
+
+def test_counter_ip_fields(api, b2b_raw_config, utils):
```
ANISH-GOTTAPU, line 4: The testname can be made inline with filename 'ipv4' instead of 'ip'

tests/utils/common.py
```diff
@@ -526,19 +526,39 @@ def get_value(field):
...
+        stack = configElement.Stack.find()[packet_header]
+    else:
+        stack = configElement.Stack.find(StackTypeId=packet_header)
     for field in stack.Field.find():
         pckt_info[field.DisplayName] = get_value(field)
     return pckt_info
 
 
 def validate_config(api, packet_header, **kwargs):
+    """
```
ANISH-GOTTAPU, line 549: We can rename it to validate_traffic_config instead of validate_config, we can discuss and do a mass update on all scripts

ashutshkumr (approved):
Merging right now as things look good as of now. @Rangababu-R please address Anish's comments in a separate PR.

Issue #122, opened 2020-12-08
'port_names' argument for result.PortRequest is ignored and results are returned for all ports


ANISH-GOTTAPU:
I have two ports `raw_tx` and `raw_rx`, and wanted the results for only one port `raw_tx`, when passed the port_names argument I am getting results for both ports 

1) Without port_names 
`    port_results = api.get_port_results(result.PortRequest())`
  dict_values([{'name': 'raw_tx', 'location': '10.39.71.139;1;1;connected', 'link': 'up', 'capture': 'stopped', 'frames_tx': 1, 'frames_rx': 0, 'frames_tx_rate': 0.0, 'frames_rx_rate': 0.0, 'bytes_tx': 74, 'bytes_rx': 0, 'bytes_tx_rate': 0.0, 'bytes_rx_rate': 0.0, 'pfc_class_0_frames_rx': 0, 'pfc_class_1_frames_rx': 0, 'pfc_class_2_frames_rx': 0, 'pfc_class_3_frames_rx': 0, 'pfc_class_4_frames_rx': 0, 'pfc_class_5_frames_rx': 0, 'pfc_class_6_frames_rx': 0, 'pfc_class_7_frames_rx': 0}, {'name': 'raw_rx', 'location': '10.39.71.139;1;2;connected', 'link': 'up', 'capture': 'stopped', 'frames_tx': 0, 'frames_rx': 1, 'frames_tx_rate': 0.0, 'frames_rx_rate': 0.0, 'bytes_tx': 0, 'bytes_rx': 74, 'bytes_tx_rate': 0.0, 'bytes_rx_rate': 0.0, 'pfc_class_0_frames_rx': 0, 'pfc_class_1_frames_rx': 0, 'pfc_class_2_frames_rx': 0, 'pfc_class_3_frames_rx': 0, 'pfc_class_4_frames_rx': 0, 'pfc_class_5_frames_rx': 0, 'pfc_class_6_frames_rx': 0, 'pfc_class_7_frames_rx': 0}])

2) With port_names argument
`    port_results = api.get_port_results(result.PortRequest(
                                       port_names=['raw_tx'])
                                        )`
[...]

PR #150, opened 2020-12-10
port_names argument support for result.PortRequest

alakendu:
This pull request for issue #122

### DIFF
--- a/ixnetwork_open_traffic_generator/vport.py
+++ b/ixnetwork_open_traffic_generator/vport.py
@@ -618,8 +618,7 @@ class Vport(object):
         
         port_filter = {'property': 'name', 'regex': '.*'}
         port_names = [port.name for port in self._api._config.ports]
-        if request is not None and request.port_names is not None \
-                and request.port_names:
+        if request and request.port_names:
             port_names = request.port_names
         if len(port_names) == 1:
             port_filter['regex'] = '^%s$' % port_names[0]
--- a/ixnetwork_open_traffic_generator/vport.py
+++ b/ixnetwork_open_traffic_generator/vport.py
@@ -615,8 +615,19 @@ class Vport(object):
             self._column_names = []
         else:
             self._column_names = request.column_names
+        
+        port_filter = {'property': 'name', 'regex': '.*'}
+        port_names = [port.name for port in self._api._config.ports]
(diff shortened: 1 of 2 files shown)


ixnetwork_open_traffic_generator/vport.py
```diff
@@ -615,8 +615,19 @@ def results(self, request):
             self._column_names = []
         else:
             self._column_names = request.column_names
+        
+        port_filter = {'property': 'name', 'regex': '.*'}
+        port_names = [port.name for port in self._api._config.ports]
+        if request is not None and request.port_names is not None \
```
ashutshkumr, line 621: If we want to be consistent with explicit checks, we should do:
```python
if request is not None and request.port_names is not None and len(request.port_names) > 0:
```

Or if we prefer something terse and implicit, we should do:
```python
if request and request.port_names:
```

ixnetwork_open_traffic_generator/vport.py
```diff
@@ -615,8 +615,19 @@ def results(self, request):
...
+                and request.port_names:
+            port_names = request.port_names
+        if len(port_names) == 1:
+            port_filter['regex'] = '^%s$' % port_names[0]
+        elif len(port_names) > 1:
+            port_filter['regex'] = '^(%s)$' % '|'.join(port_names)
+
```
ashutshkumr, line 628: Shouldn't just executing the snippet under `else` (for regex) work for all cases ?

ashutshkumr (approved):
Apart from comments on style, LGTM.

ajbalogh (approved):

ixnetwork_open_traffic_generator/vport.py
alakendu: Actually we coded that way though I don't have any preference. 
Note: Zen of Python said "Explicit is better than implicit".  :)

ixnetwork_open_traffic_generator/vport.py
alakendu: we will ignore corner case when len(port_names) == 0. So I still vote for elif.

ixnetwork_open_traffic_generator/vport.py
alakendu: add implicit check

PR #156, opened 2020-12-11
Remove duplicate dscp phb entry

ajbalogh:
Duplicate entry causes incorrect stack/field entry.

### DIFF
--- a/ixnetwork_open_traffic_generator/customfield.py
+++ b/ixnetwork_open_traffic_generator/customfield.py
@@ -22,7 +22,6 @@ class CustomField(object):
         '26': 'ipv4.header.priority.ds.phb.assuredForwardingPHB.assuredForwardingPHB',
         '28': 'ipv4.header.priority.ds.phb.assuredForwardingPHB.assuredForwardingPHB',
         '30': 'ipv4.header.priority.ds.phb.assuredForwardingPHB.assuredForwardingPHB',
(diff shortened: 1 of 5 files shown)


ixnetwork_open_traffic_generator/customfield.py
```diff
@@ -22,7 +22,6 @@ class CustomField(object):
         '26': 'ipv4.header.priority.ds.phb.assuredForwardingPHB.assuredForwardingPHB',
         '28': 'ipv4.header.priority.ds.phb.assuredForwardingPHB.assuredForwardingPHB',
         '30': 'ipv4.header.priority.ds.phb.assuredForwardingPHB.assuredForwardingPHB',
-        '24': 'ipv4.header.priority.ds.phb.assuredForwardingPHB.assuredForwardingPHB',
```
alakendu, line 25: I think we should replace 24 with 34 rather than remove it. Also update model ( Dscp ) accordingly as both PHB_CS3 and PHB_AF41 map with '24'
```python
PHB_CS3 = '24'
PHB_AF41 = '24'
```
According to rfc2597
   The RECOMMENDED values of the AF codepoints are as follows: AF11 = '
   001010', AF12 = '001100', AF13 = '001110', AF21 = '010010', AF22 = '
   010100', AF23 = '010110', AF31 = '011010', AF32 = '011100', AF33 = '
   011110', AF41 = '100010', AF42 = '100100', and AF43 = '100110'.  The
   table below summarizes the recommended AF codepoint values.

alakendu (approved):

Issue #161, opened 2020-12-14
Traffic transmit state shows always 'started' for fixed flow even though the flow stops after transmitting all packets if there is other flow(continuous) in started state


ANISH-GOTTAPU:
Scenario - 1
Consider two flows f1 & f2 
1) f1 with fixed packet count
2) f2 with fixed packet count
<b>Transmit state once all packets are transmitted
stopped
stopped</b>

Scenario - 2
Consider two flows f1 & f2 
1) f1 with fixed packet count
2) f2 with continuous packet count
<b>Transmit state once packets are transmitted for f1
started
started</b>

<b>Note</b>: The traffic is not stopped intentionally in any Scenario

I assume, in scenario 2 the f1 state should be in stopped instead of started state as all packets are transmitted. When I dig deeper the states are coming from REST API, raising an issue as a counter issue for any customer use case

ajbalogh:
This is a bug in ixnetwork middleware and can be reproduced by using the full UI and creating two traffic items, one continuous, the other stopping after a predefined number of packets.  This package will have to workaround that issue when returning flow stats. There is currently code in there to check the Tx Frame Rate and return started if it is greater than 0.  That section should be expanded to check the configelement rate and adjust the state based on frames tx etc.
At the same time a bug should be opened with ixnetwork traffic but the customer cannot wait for an update/hotfix/patch etc.
This issue is being seen by the current sonic use cases.

ANISH-GOTTAPU:
test script
[test_traffic_transmit_state.zip](https://github.com/open-traffic-generator/ixnetwork/files/5696562/test_traffic_transmit_state.zip)

ashutshkumr:
Verfied using test_traffic_transmit_state.py

Issue #160, opened 2020-12-14
'NOVUS10/5/2.5/1/100M16DP' LM always defaults to Aggregation mode with "speed_10_gbps"


ANISH-GOTTAPU:
'NOVUS10/5/2.5/1/100M16DP' LM always defaults to 10G Aggregated mode if speed is set to "speed_10_gbps"

From GUI there are two modes as shown
![image](https://user-images.githubusercontent.com/48308607/102049953-e30e7200-3e07-11eb-83e2-81e4c579b7a8.png)

But the restpy return the available modes as ['normal', 'tenGigAggregation'] which are mapped to '10G Non-agg'&'10G agg' respectively

<b>RootCause</b>:
`available_mode` is matched with card available modes and `speed_10_gpbs` is matching with `tenGigAggregation` , so always the card ends up in tenGigAggregation. 

code block ref:
https://github.com/open-traffic-generator/ixnetwork/blob/master/ixnetwork_open_traffic_generator/vport.py#L416

ashutshkumr:
fixed as per `snappi[ixnetwork]==0.3.13`

Issue #155, opened 2020-12-11
If physical port speed and config speed is same, do we need to reconfigure it?


Rangababu-R:
if the port was set to 100g already by manually or automation. 
OTG is not picking that info and applying the speed on all ports and leading to the issue #153

do we need to reconfigure it?

ashutshkumr:
The problem is, it's still not clear what we should do if the speed is not provided by user.
Picking up a fixed default speed (e.g. 1G) is not an option here. We might need to resort to traffic engine specific default.

This needs more discussion.
@ajbalogh @ankur-sheth @winstonliu1111 @alakendu

ashutshkumr:
fixed as per `snappi[ixnetwork]==0.3.13`

PR #178, opened 2020-12-17
Novus 10G and 100G

alakendu:
This will address #155 #160 and #161

### DIFF
--- a/ixnetwork_open_traffic_generator/ixnetworkapi.py
+++ b/ixnetwork_open_traffic_generator/ixnetworkapi.py
@@ -347,6 +347,42 @@ class IxNetworkApi(Api):
     def _get_topology_name(self, port_name):
         return 'Topology %s' % port_name
 
+    def select_card_aggregation(self, location):
+        (hostname, cardid, portid) = location.split(';')
+        payload = {
+            'selects': [{
+                'from':
+                '/availableHardware',
+                'properties': [],
+                'children': [{
+                    'child':
+                    'chassis',
+                    'properties': [],
+                    'filters': [{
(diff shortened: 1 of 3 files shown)


alakendu:
I am trying to fix probable use case for Novus10G and Novus100G load module without interrupting existing code block. I feel still their are scope of improvement and overcome collateral. Please review this pull request.

ajbalogh (approved):

ANISH-GOTTAPU:
Hi Alakendu,

The media type is not set to fiber and always defaults to copper for 10G card with the updated code

Please find the test attached to verify 

[test_layer1_speed_Novus_10G.zip](https://github.com/open-traffic-generator/ixnetwork/files/5714424/test_layer1_speed_Novus_10G.zip)

ANISH-GOTTAPU:
@alakendu 

For 100G Novus, 
1) updating two ports from 100G-40G or vice versa is taking nearly 4 minutes, not sure this can be optimized as the RG is updating.
2) updating two ports from same speed to same speed is taking 2 minutes, I feel this can be optimized by checking the location and speed and don't do any operation at port level. You can put a thought to it

ANISH-GOTTAPU:
@alakendu : If I am running the script `test_layer1_speed_Novus_10G.py` of PR #194 auto negotiation is not applied. If I run separately for 10G it is working but not working if i execute in loop with different speeds. please check once

ashutshkumr (approved):
There are other outstanding issues that will be fixed in separate PRs. Merging this for now since changes look good and there wasn't any collateral reported.

Issue #174, opened 2020-12-16
Mutltiple testcases with diff flows are failing


ANISH-GOTTAPU:
If a user is doing two tests `test1` with flows f1&f2 and `test2` the user only wants to use f1 but f2 is still intact in the ixnetwork and effects the test validation

Steps to produce:
test1:
`tests/statistics/test_port_and_flow_stats.py`
test2:
`tests/traffic/test_tcp_flow_capture.py`

Both tests are passing if run separately, but if run in sequential order, test2 would fail as it is holding up flow f2 from test1

alakendu:
I am not able to reproduced it in latest code in my own test script. Actually test1 (tests/statistics/test_port_and_flow_stats.py) not available within head of master branch.
Could you please double check and let me know the outcome as well as script of test1

ANISH-GOTTAPU:
@alakendu : The script was still part of PR and is not merged yet. Attaching the script here
[test_port_and_flow_stats.zip](https://github.com/open-traffic-generator/ixnetwork/files/5722363/test_port_and_flow_stats.zip)

alakendu:
Outcome after
test1: (two flows f1 and f2 are configured)
tests/statistics/test_port_and_flow_stats.py
![image](https://user-images.githubusercontent.com/35574510/102772719-ff815000-43ad-11eb-9feb-25fafe2e7822.png)

test2: (Only f1 is configured)
tests/traffic/test_tcp_flow_capture.py
![image](https://user-images.githubusercontent.com/35574510/102772815-2ccdfe00-43ae-11eb-9520-4ef32b0b4db6.png)

I think those are configured according to the expectation. Please let me know if I have missed something

ANISH-GOTTAPU:
please use this branch `issue174` and execute the tests with pytest mark issue174

`python -m pytest tests -m "issue174"`

alakendu:
According to our discussion this is happening due to manipulation in common flow property
```python
b2b_raw_config.flows.append(flow2)
```

Please handle it within test case.

ANISH-GOTTAPU:
Addressed with PR #190

PR #187, opened 2020-12-21
tests to reproduce #174

ANISH-GOTTAPU:
Don't commit this branch. This is just to reproduce issue 174

The issue is reproduced with tests run in a particular order. The order of tests is added with mark "issue174". please use below cli command to reproduce the issue

`python -m pytest tests -m "issue174"`


ANISH-GOTTAPU:
root cause identified. closing the PR

PR #188, opened 2020-12-21
updating stat tests

ANISH-GOTTAPU:
updating stat tests


ANISH-GOTTAPU:
closing PR as it is addressed in #190

PR #190, opened 2020-12-22
More infra changes

ashutshkumr:
- Make sure utils/settings/b2bconfig fixtures are scoped to single test script
- Relocate stats tests to traffic/ because that dir contains (or will contain) TCs related to stats/scheduler which require actual traffic operation
- Removed some old TCs for which I think we already have the coverage

@ANISH-GOTTAPU needs to further investigate why we're seeing failure in stats filter unit test.


ashutshkumr:
Had a discussion with @alakendu and @ANISH-GOTTAPU . There's an issue with header retention in one of the test scripts which has been skipped for now.

ANISH-GOTTAPU (approved):

Issue #191, opened 2020-12-22
Header retained in the flow from previous run


ANISH-GOTTAPU:
`test1`: Add tcp header to the flow `f1`

`test2`: Set the config with raw ethernet header for flow `f1`

Expectation: The flow `f1` should have only raw ethernet header but tcp header is retained from `test1`

`test1`: pytest tests/tcp/test_tcp_header_with_fixed.py
`test2`: pytest tests/traffic/test_stats_filter.py

ashutshkumr:
fixed as per `snappi[ixnetwork]==0.3.13`

PR #195, opened 2020-12-22
Header retained in the flow from previous run

alakendu:
fix for #191

### DIFF
--- a/ixnetwork_open_traffic_generator/trafficitem.py
+++ b/ixnetwork_open_traffic_generator/trafficitem.py
@@ -27,7 +27,7 @@ class TrafficItem(CustomField):
 
     _TYPE_TO_HEADER = {
         'ethernet': 'ethernet',
-        'pfcPause': 'pfc_pause',
+        'pfcPause': 'pfcpause',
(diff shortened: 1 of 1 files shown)


ajbalogh (commented):
Rather than check the length of headers would a better check be to compare header choice vs stacktypeid - if its not a direct match then just recreate the stack using the headers rather than attempting to modify?

alakendu:
> Rather than check the length of headers would a better check be to compare header choice vs stacktypeid - if its not a direct match then just recreate the stack using the headers rather than attempting to modify?

@ajbalogh At the time of testing I encounter some more issue in our existing Remove() mechanism.  (say Test1 : Ether, IP, Tcp and Test2 : pfcPause, Udp Result: pfcPause, IP, Tcp) It looks like IxNetwork is doing some serialization for newly created stack. Therefore I am not getting predictive result.  So I have recheck newly created stacks and remove accordingly. I know it will good if we did entire operation in one go. Could you please review and suggest any better solution.

ajbalogh (approved):

ashutshkumr (approved):

PR #198, opened 2020-12-23
added script for measuring api call duration

ashutshkumr:
Currently only single and multiple flow use cases have been added. More maybe added in future based on requirements.

Output:

|config_type      |iterations|set_config_ms|start_flows_ms|stop_flows_ms|start_capture_ms|stop_capture_ms|get_port_results_ms|get_flow_results_ms|get_captures_ms|result|
|-----------------|----------|-------------|--------------|-------------|----------------|---------------|-------------------|-------------------|---------------|------|
|one_flow_tcp_128b|2         |64019        |27873         |516          |0               |0              |2782               |2815               |37466          |PASSED|
|ten_flows_tcp_128b|2         |254117       |32805         |579          |0               |0              |2700               |2711               |37698          |PASSED|

### DIFF
--- a/tests/traffic/test_api_perf.py
+++ b/tests/traffic/test_api_perf.py
@@ -1,0 +1,244 @@
+import pytest
+import utils
+import traceback
+import time
+from abstract_open_traffic_generator import flow, control, result
+
+
+CSV_COLUMNS = [
+    'config_type', 'iterations', 'set_config_ms', 'start_flows_ms',
+    'stop_flows_ms', 'start_capture_ms', 'stop_capture_ms',
+    'get_port_results_ms', 'get_flow_results_ms', 'get_captures_ms', 'result'
+]
+CSV_DIR = utils.new_logs_dir()
+
+
+@pytest.mark.e2e
+@pytest.mark.parametrize(
+    'config_type', ['one_flow_tcp_128b', 'ten_flows_tcp_128b']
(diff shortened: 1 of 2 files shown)


tests/traffic/test_api_perf.py
```diff
@@ -0,0 +1,244 @@
...
+@pytest.mark.parametrize("iterations", [10])
+def test_api_perf(settings, config_type, iterations):
+    """
+    This test collects avg duration in milliseconds for all required API calls
+    required in a usual E2E test script.
+    """
+    packets = 100
+    row = new_result_dict(config_type, iterations)
+
+    for i in range(1, iterations + 1):
```
Rangababu-R, line 29: as "iterations" is parametrize variable, instead of using for loop can leverage the iteration mechanism of pytest.

@pytest.mark.parametrize("iterations", list(range(10)))

ashutshkumr: I agree, the only problem is that we'd have to think about managing snippets living outside the loop. e.g L57-60. Any suggestions on that ?

Rangababu-R: can be made that as fixture and scope it, but it doesn't make sense just for one script to do so.
if these tests grow can think of it later.

Rangababu-R (approved):

Issue #203, opened 2021-01-05
AutoNegotiation is not applied on 10G Novus if previous setting is intact with 1G


ANISH-GOTTAPU:
AutoNegotiation is not applied on 10G Novus card and speed  of 10G is not applied because of this. This issue occurs only if existing config is intact with 1G AutoNegotation applied

Please run the script attached to reproduce the issue
[test_layer1_speed_Novus_10G.zip](https://github.com/open-traffic-generator/ixnetwork/files/5769759/test_layer1_speed_Novus_10G.zip)

PR #210, opened 2021-01-07
AutoNegotiation is not applied on 10G Novus

alakendu:
This should address this fix https://github.com/open-traffic-generator/ixnetwork/issues/203

### DIFF
--- a/ixnetwork_open_traffic_generator/vport.py
+++ b/ixnetwork_open_traffic_generator/vport.py
@@ -608,6 +608,9 @@ class Vport(object):
     def _set_gigabit_auto_negotiation(self, vport, layer1, imports):
         advertise = []
         advertise.append(Vport._SPEED_MAP[layer1.speed])
+        auto_field_name  = 'enableAutoNegotiation'
+        if re.search('novustengiglan', vport['type'].lower()) is not None:
+            auto_field_name = 'autoNegotiate'
         proposed_import = {
             'xpath':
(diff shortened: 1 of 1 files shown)


ixnetwork_open_traffic_generator/vport.py
```diff
@@ -591,29 +591,40 @@ def _set_ethernet_auto_negotiation(self, vport, layer1, imports):
...
+        advertise.append(Vport._SPEED_MAP[layer1.speed])
         proposed_import = {
             'xpath':
             vport['xpath'] + '/l1Config/' + vport['type'].replace('Fcoe', ''),
             'ieeeL1Defaults':
             layer1.ieee_media_defaults,
             'speed':
             Vport._SPEED_MAP[layer1.speed],
-            'enableAutoNegotiation':
+            'autoNegotiate':
```
ajbalogh, line 618: The name of this property is conditional depending on the l1Config -currentType. This method should be expanded to set property names based on the l1Config -currentType.

ajbalogh (changes requested):
Rather than remove attributes have multiple methods setting attributes based on l1Config -currentType. A good way to get import properties is here: https://openixia.github.io/ixnetwork_openapi/

ixnetwork_open_traffic_generator/vport.py
alakendu: Add validation https://github.com/open-traffic-generator/ixnetwork/pull/210/commits/c4d8c89eb8ace2c13c4c66b6e534859efbe3af41

Please review

Issue #216, opened 2021-01-08
Create Test for ipv4 Device config with traffic and capture test


Rangababu-R:
1. Create ipv4 device config on Tx and Rx ports.
2. validate the device configs.
3. Create traffic with ipv4 device names as end points
4. enable traffic
5. run traffic

Validation:
Validate the source and destination, mac and ipv4 addresses with in the captures.

ashutshkumr:
fixed as per `snappi[ixnetwork]==0.3.13`

PR #218, opened 2021-01-08
Test for device Config and counter pattern

Rangababu-R:
Test for device donfig and counter pattern
#216


ashutshkumr (changes requested):
- relocate this test to `traffic` directory
- configure two flows instead of one (differentiated by size)
- packet description should be empty for first flow
- second flow should have headers Ether() / IP() / UDP(counter-source-port)
- validate ip and mac addres for first flow and ip, mac, udp port for second flow

Rangababu-R:
will create new PR. I feel the branch is not clean

PR #221, opened 2021-01-11
adding layer1 tests

ANISH-GOTTAPU:
adding layer1 tests

### DIFF
--- a/tests/layer1/test_layer1_flow_control_8021qbb.py
+++ b/tests/layer1/test_layer1_flow_control_8021qbb.py
@@ -1,0 +1,66 @@
+from abstract_open_traffic_generator.config import *
+from abstract_open_traffic_generator.layer1 import *
+from abstract_open_traffic_generator.control import *
+
+
+def test_layer1_flow_control_8021qbb(api, tx_port, rx_port, options, utils):
+    """
+    Test that layer1 flow control 8021qbb configuration settings
+    are being applied correctly.
+
+    Validation: Validate the layer1 8021qbb properties applied using Restpy
+    """
+    port1_delay = 3
+    port1_pfc_priority_groups = [1, 0, -1, -1, 7, -1, -1, -1]
+    directed_address = '01 80 C2 00 00 01'
+
+    enabled_pfc = Ieee8021qbb(pfc_delay=3,
+                              pfc_class_0=port1_pfc_priority_groups[0],
+                              pfc_class_1=port1_pfc_priority_groups[1],
+                              pfc_class_4=port1_pfc_priority_groups[4])
+    fcoe1 = Layer1(name='enabled pfc delay',
+                   port_names=[tx_port.name],
+                   speed=utils.settings.speed,
+                   auto_negotiate=True,
+                   media='fiber',
+                   flow_control=FlowControl(directed_address=directed_address,
+                                            choice=enabled_pfc))
+
+    disabled_pfc = Ieee8021qbb(pfc_delay=0)
+    fcoe2 = Layer1(name='disabled pfc delay',
+                   port_names=[rx_port.name],
+                   auto_negotiate=True,
+                   speed=utils.settings.speed,
+                   media='fiber',
+                   flow_control=FlowControl(directed_address=directed_address,
+                                            choice=disabled_pfc))
+
+    config = Config(ports=[tx_port, rx_port],
+                    layer1=[fcoe1, fcoe2],
+                    options=options)
+    api.set_state(State(ConfigState(config=config, state='set')))
+    validate_8021qbb_config(api,
+                            port1_delay,
+                            port1_pfc_priority_groups,
+                            directed_address)
+
+
+def validate_8021qbb_config(api,
+                            port1_delay,
+                            port1_pfc_priority_groups,
+                            directed_address):
+    """
+    Validate 8021qbb config using Restpy
+    """
+    ixnetwork = api._ixnetwork
+    port1 = ixnetwork.Vport.find()[0]
+    port2 = ixnetwork.Vport.find()[1]
+    type = port1.Type.replace('Fcoe', '')
+    type = type[0].upper() + type[1:]
+    port1_type = eval('port1.L1Config.' + type)
+    port1_fcoe = (eval('port1.L1Config.' + type + '.Fcoe'))
+    port2_fcoe = (eval('port2.L1Config.' + type + '.Fcoe'))
+    assert port1_type.FlowControlDirectedAddress == directed_address
+    assert port1_fcoe.PfcPauseDelay == port1_delay
+    assert port1_fcoe.PfcPriorityGroups == port1_pfc_priority_groups
+    assert port2_fcoe.EnablePFCPauseDelay is False
--- a/tests/layer1/test_layer1_flow_control_8023x.py
+++ b/tests/layer1/test_layer1_flow_control_8023x.py
@@ -1,0 +1,56 @@
+from abstract_open_traffic_generator.config import *
+from abstract_open_traffic_generator.layer1 import *
+from abstract_open_traffic_generator.control import *
+
+
+def test_layer1_flow_control_8023x(api, tx_port, rx_port, options, utils):
+    """
+    Test that layer1 flow controle 8023x configuration settings
+    are being applied correctly.
+
+    Validation: Validate the layer1 properties applied using Restpy
+    """
+
+    directed_address = '01 80 C2 00 00 01'
+
+    enabled_pfc = Ieee8023x()
+    fcoe1 = Layer1(name='pfc delay-1',
+                   port_names=[tx_port.name],
+                   speed=utils.settings.speed,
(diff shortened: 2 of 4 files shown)


tests/layer1/test_layer1_flow_control_8021qbb.py
```diff
@@ -0,0 +1,66 @@
...
+
+    enabled_pfc = Ieee8021qbb(pfc_delay=3,
+                              pfc_class_0=port1_pfc_priority_groups[0],
+                              pfc_class_1=port1_pfc_priority_groups[1],
+                              pfc_class_4=port1_pfc_priority_groups[4])
+    fcoe1 = Layer1(name='enabled pfc delay',
+                   port_names=[tx_port.name],
+                   speed=utils.settings.speed,
+                   auto_negotiate=True,
+                   media='fiber',
```
ashutshkumr, line 25: use media from settings

tests/layer1/test_layer1_flow_control_8021qbb.py
```diff
@@ -0,0 +1,66 @@
...
+                            directed_address):
+    """
+    Validate 8021qbb config using Restpy
+    """
+    ixnetwork = api._ixnetwork
+    port1 = ixnetwork.Vport.find()[0]
+    port2 = ixnetwork.Vport.find()[1]
+    type = port1.Type.replace('Fcoe', '')
+    type = type[0].upper() + type[1:]
+    port1_type = eval('port1.L1Config.' + type)
```
ashutshkumr, line 60: why are we using eval here ? I think these can be achieved using `getattr` or `setattr`

tests/layer1/test_layer1_flow_control_8023x.py
```diff
@@ -0,0 +1,56 @@
+from abstract_open_traffic_generator.config import *
+from abstract_open_traffic_generator.layer1 import *
+from abstract_open_traffic_generator.control import *
+
```
ashutshkumr, line 4: please incorporate changes mentioned in 8021qbb test here as well

tests/layer1/test_layer1_speed_Novus_100G.py
```diff
@@ -0,0 +1,86 @@
...
+
+    Validation: Validate the layer1 properties applied using Restpy
+    """
+    speed_type = {'speed_10_gbps': '10000',
+                  'speed_25_gbps': '25000',
+                  'speed_40_gbps': '40000',
+                  'speed_100_gbps': '100000'}
+    port1 = Port(name='port1', location=utils.settings.ports[0])
+    port2 = Port(name='port2', location=utils.settings.ports[1])
+    speed = utils.settings.speed
```
ashutshkumr, line 25: I think we'll disregard speeds fetched from settings

tests/layer1/test_layer1_speed_Novus_100G.py
```diff
@@ -0,0 +1,86 @@
+import pytest
+from abstract_open_traffic_generator.port import *
+from abstract_open_traffic_generator.config import *
+from abstract_open_traffic_generator.layer1 import *
+from abstract_open_traffic_generator.control import *
+
+
```
ashutshkumr, line 7: we should ensure that this test doesn't run unless we provide a specific tag. e.g. if "e2e" or "not e2e" tags are provided, do not run this test.

tests/layer1/test_layer1_speed_Novus_100G.py
```diff
@@ -0,0 +1,86 @@
...
+    """Test that layer1 configuration settings are being applied correctly
+    A user should be able to configure ports with/without locations.
+    The expectation should be if a location is configured the user wants to
+    connect but debug should allow for config creation without location.
+    Ports with no location should not generate an error message.
+    Ports with location should generate an error message if unable to connect.
+
+    Validation: Validate the layer1 properties applied using Restpy
+    """
+    speed_type = {'speed_10_gbps': '10000',
```
ashutshkumr, line 19: I know speed 50G is not included in models, and hence would suggest parametrizing the test with supported speeds (so we can opt out of any given speed).

tests/layer1/test_layer1_speed_Novus_100G.py
```diff
@@ -0,0 +1,86 @@
...
+                  'speed_25_gbps': '25000',
+                  'speed_40_gbps': '40000',
+                  'speed_100_gbps': '100000'}
+    port1 = Port(name='port1', location=utils.settings.ports[0])
+    port2 = Port(name='port2', location=utils.settings.ports[1])
+    speed = utils.settings.speed
+    auto_negotiate = True
+    ieee_media_defaults = False
```
ashutshkumr, line 27: there should be at least one use case where ieee_media_defaults should be set to True. In such cases, one need not pass following:
- ieee_media_defaults
- auto_negotiate
- AutoNegotiation

tests/layer1/test_layer1_speed_Novus_100G.py
```diff
@@ -0,0 +1,86 @@
...
+    Ports with no location should not generate an error message.
+    Ports with location should generate an error message if unable to connect.
+
+    Validation: Validate the layer1 properties applied using Restpy
+    """
+    speed_type = {'speed_10_gbps': '10000',
+                  'speed_25_gbps': '25000',
+                  'speed_40_gbps': '40000',
+                  'speed_100_gbps': '100000'}
+    port1 = Port(name='port1', location=utils.settings.ports[0])
```
ashutshkumr, line 23: we should only extract card number and chassis ip from port location. Since we need to run test on all speeds, port number should be assigned to them accordingly for each speed.

tests/layer1/test_layer1_speed_Novus_100G.py
```diff
@@ -0,0 +1,86 @@
...
+    ixnetwork = api._ixnetwork
+    port1 = ixnetwork.Vport.find()[0]
+    port2 = ixnetwork.Vport.find()[1]
+    type = port1.Type
+    type = type[0].upper() + type[1:]
+    ports_assigned = [vport.Location for vport in ixnetwork.Vport.find()]
+    assert (all([vport.ActualSpeed == int(speed_type[speed])
+                for vport in ixnetwork.Vport.find()]))
+    assert ports_assigned == ports
+    assert (eval(
```
ashutshkumr, line 74: replace eval with setattr or getattr

tests/layer1/test_layer1_speed_Novus_10G.py
```diff
@@ -0,0 +1,72 @@
+import pytest
+from abstract_open_traffic_generator.port import *
+from abstract_open_traffic_generator.config import *
+from abstract_open_traffic_generator.layer1 import *
+from abstract_open_traffic_generator.control import *
+
```
ashutshkumr, line 6: most changes requested in Novus100G script applies here as well.

ashutshkumr (changes requested):
Please implement all requested changes and let me know if there's any concern.

tests/layer1/test_layer1_speed_Novus_100G.py
ANISH-GOTTAPU: didn't get the comment. "e2e" tag is already added, are you expecting to add a reason?

tests/layer1/test_layer1_speed_Novus_100G.py
ANISH-GOTTAPU: The dictionary is not testing for each speed. The dict is used for validation. 

ports and speed will be taken from settings file
    "ports": [
        "10.36.87.215;1;3",
        "10.36.87.215;1;4"
    ],
    "speed": "speed_100_gbps",

The port location and speed should be user's input as 40G&100G takes same port location

tests/layer1/test_layer1_speed_Novus_100G.py
ANISH-GOTTAPU: I assume we haven't discussed about script should get card number and chassis ip from port location. If port location and speed from the user is not matching then OTG should throw an error. We can have a discussion tomorrow to conclude on this

tests/layer1/test_layer1_speed_Novus_100G.py
ANISH-GOTTAPU: I think still the user should send speed, as 40G&100G takes same ports and the case is same for 10G&25G. We can't conclude speed based on port locations.

tests/layer1/test_layer1_speed_Novus_100G.py
ANISH-GOTTAPU: `ieee_media_defaults` is True as per the model. I felt this is covered in all other tests, we can discuss tomorrow and conclude

tests/layer1/test_layer1_speed_Novus_10G.py
ANISH-GOTTAPU: updated

ashutshkumr (approved):

Issue #235, opened 2021-01-14
Flow framesize is being ignored


ajbalogh:
User configures multiple flows on a single port one of which contains only a pfcpause packet header with a framesize of 64 bytes.
The flow that contains the pfcpause packet header has its framesize adjusted to a larger size.

The configuration should be immutable (ie if the user asks for 64 byte framesize it should provide that or return an error indicating it is unable to do that).

This issue is being seen with the following test when being run against mellanox 2700:
https://github.com/Azure/sonic-mgmt/blob/master/tests/pfc/test_pfc_pause_lossy.py

ashutshkumr:
https://github.com/open-traffic-generator/ixnetwork/releases/tag/v0.0.73

PR #236, opened 2021-01-15
Flow framesize is being ignored #235

alakendu:
This should address https://github.com/open-traffic-generator/ixnetwork/issues/235

Please take a look and let me know if this will address customer use case as well as do not regress anything

### DIFF
--- a/ixnetwork_open_traffic_generator/trafficitem.py
+++ b/ixnetwork_open_traffic_generator/trafficitem.py
@@ -204,7 +204,7 @@ class TrafficItem(CustomField):
                 self._configure_size(ixn_ce, flow.size)
                 self._configure_rate(ixn_ce, flow.rate)
                 self._configure_tx_control(ixn_ce, hl_stream_count, flow.duration)
-                self._configure_options(flow)
+            self._configure_options()
 
     def _configure_tracking(self, flow, ixn_tracking):
         """Set tracking options"""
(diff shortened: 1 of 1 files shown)


ashutshkumr (approved):
@ajbalogh FYI, this fixes PFC-specific use cases. Following pending items will be addressed in separate PR:
- Expose warnings in response
- Conditionally enable min frame size (use 4-byte signature) agnostic of underlying headers

@Rangababu-R kindly update on validation when done.

Rangababu-R (approved):

ixnetwork_open_traffic_generator/trafficitem.py
```diff
@@ -215,12 +215,13 @@ def _configure_tracking(self, flow, ixn_tracking):
...
         
-    def _configure_options(self, flow):
+    def _configure_options(self):
         enable_min_frame_size = False
-        if (len(flow.packet) == 1 and flow.packet[0].choice == 'pfcpause'
-                and flow.size is not None
-                and flow.size.choice == 'fixed' and flow.size.fixed <= 64):
-            enable_min_frame_size = True
+        for flow in self._api.config.flows:
+            if (len(flow.packet) == 1 and flow.packet[
```
ajbalogh, line 221: This will not address the scenario where there are multiple flows, one of which is a pfc pause flow (what the customer is seeing).
It should be if there is a /globals/apperror/errors that is 'Frame Size Adjusted' and a pfc pause flow then set the min frame size regardless of other flows.

ajbalogh (changes requested):

ixnetwork_open_traffic_generator/trafficitem.py
ashutshkumr: @ajbalogh, could you please clarify why you think it won't work for multiple flows ? These changes were done to exclusively take into account multiple flows (with at least one pfc pause flow).

@Rangababu-R could you please confirm if following test worked as expected after changing frame size in pause flow from 128B to 64B (I think you already did, just need a confirmation) ?
- https://github.com/open-traffic-generator/ixnetwork/blob/master/tests/pfc/test_pfc_pause_e2e.py

ajbalogh (approved):

PR #238, opened 2021-01-18
Support RequestDetail

alakendu:
This fix will return RequestDetail object for set_state. So the code looks like:
    config_state = control.ConfigState(config=config, state='set')
    result = api.set_state(control.State(config_state))
    print(result.errors)
    print(result.warnings)

Please review and let me know if this will fulfill the requirement.


ixnetwork_open_traffic_generator/ixnetworkapi.py
```diff
@@ -290,6 +315,12 @@ def _connect(self):
                 version = "ixnetwork-open-traffic-generator not installed " \
                     "using pip, unable to determine version"
             self.info(version)
+        self._backup_erros()
```
ajbalogh, line 318: typo - instances of erros should be errors

ashutshkumr (approved):

PR #251, opened 2021-02-04
Details support

alakendu:


alakendu:
Ignore lgtm-com  check as those are diff with main for previous change. Those will address when snappi_ixnetwork will merge with main.

PR #252, opened 2021-02-04
metrics fix

### LINKED ISSUE #250: api.get_metrics should return MetricsResponse instead of any one of its child

Rangababu-R:
bug fix #250

### DIFF
--- a/snappi_ixnetwork/ixnetworkapi.py
+++ b/snappi_ixnetwork/ixnetworkapi.py
@@ -1,5 +1,4 @@
 import json
-import pdb
 import time
 from collections import namedtuple
 from ixnetwork_restpy import TestPlatform, SessionAssistant
@@ -219,8 +218,11 @@ class Api(snappi.Api):
             metric_res = self.metrics_response()
             metric_res.flow_metrics.deserialize(response)
             return metric_res
-        if request.choice == 'bgpv4':
-            return
+        if request.choice is not None:
+            msg = "{} is currently not supported \
+                regret for the inconvenience \
+                please check for latest release".format(request.choice)
+            raise Exception(msg)
 
     def add_error(self, error):
         """Add an error to the global errors
--- a/snappi_ixnetwork/ixnetworkapi.py
+++ b/snappi_ixnetwork/ixnetworkapi.py
@@ -219,9 +219,9 @@ class Api(snappi.Api):
             metric_res.flow_metrics.deserialize(response)
             return metric_res
         if request.choice is not None:
-            msg = "{} is currently not supported \
-                regret for the inconvenience \
-                please check for latest release".format(request.choice)
+            msg = "{} is not a supported choice for metrics; \
+            the supported choices are \
+            ['port', 'flow']".format(request.choice)
             raise Exception(msg)
 
     def add_error(self, error):
--- a/snappi_ixnetwork/ixnetworkapi.py
+++ b/snappi_ixnetwork/ixnetworkapi.py
@@ -1,4 +1,5 @@
 import json
+import pdb
 import time
 from collections import namedtuple
 from ixnetwork_restpy import TestPlatform, SessionAssistant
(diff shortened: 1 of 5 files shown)


snappi_ixnetwork/ixnetworkapi.py
```diff
@@ -1,4 +1,5 @@
 import json
+import pdb
```
ashutshkumr, line 2: please remove `pdb`

snappi_ixnetwork/ixnetworkapi.py
```diff
@@ -210,12 +211,14 @@ def get_metrics(self, request):
...
+            return metric_res
         if request.choice == 'flow':
             response = self.traffic_item.results(request.flow)
-            return self.metrics_response().flow_metrics.\
-                deserialize(response)
+            metric_res = self.metrics_response()
+            metric_res.flow_metrics.deserialize(response)
+            return metric_res
         if request.choice == 'bgpv4':
             return
```
ashutshkumr, line 223: Please raise a user-friendly error for unsupported metric choices (listing down the supported ones)

ashutshkumr (changes requested):
please address the minor changes mentioned.

snappi_ixnetwork/ixnetworkapi.py
```diff
@@ -210,14 +210,19 @@ def get_metrics(self, request):
         # Need to change the code style when the choice Enum grows big
         if request.choice == 'port':
             response = self.vport.results(request.port)
-            return self.metrics_response().port_metrics.\
-                deserialize(response)
+            metric_res = self.metrics_response()
+            metric_res.port_metrics.deserialize(response)
+            return metric_res
```
alakendu, line 215: better use elif as we are decide according to choice

snappi_ixnetwork/trafficitem.py
```diff
@@ -610,7 +610,7 @@ def results(self, request):
         """Return flow results
         """
         # setup parameters
-        self._column_names =  request._properties.get('column_names')
+        self._column_names = request._properties.get('column_names')
```
alakendu, line 613: _properties is really require. We can access that through request.column_names

snappi_ixnetwork/vport.py
```diff
@@ -698,8 +698,9 @@ def results(self, request):
             port_row = dict()
             self._set_result_value(port_row, 'name', vport.get('name'))
             location = vport.get('location')
-            if vport.get('connectionState').\
-                startswith('connectedLink') is True:
+            if vport.get(
+                'connectionState'
+            ).startswith('connectedLink') is True:
                 location += ';connected'
             elif len(location) > 0:
```
alakendu, line 705: put this code within exception handling block
self._set_result_value(port_row, ext_name, row_val, typ)

alakendu (changes requested):

snappi_ixnetwork/trafficitem.py
Rangababu-R: true but if the column_names are not set from the script then request.column_names will create an instance (as it is @property decorator), to avoid used _properties

ashutshkumr: @alakendu @Rangababu-R let's address this in a separate PR. for now it's not that harmful

snappi_ixnetwork/vport.py
Rangababu-R: I did not get it. This is old code I just indented as per flake8.
point me to the detail

snappi_ixnetwork/ixnetworkapi.py
Rangababu-R: true but, choice can hold anyone of the metric right. 
from my thought elif would be needed when the choice holds multiple values. correct me if I am wrong

snappi_ixnetwork/vport.py
alakendu: I am talking about this code which present earlier:

```python
                    for ext_name, int_name, typ in self._RESULT_COLUMNS:
                        try:
                            self._set_result_value(port_row, ext_name,
                                                   row[int_name], typ)
                        except Exception:
                            # TODO print a warning maybe ?
                            pass
```

snappi_ixnetwork/trafficitem.py
alakendu: Yes this is just a concept. You can take any one. 

@Rangababu-R  I feel that will create because it set within constructor
```python
    def __init__(self, parent=None, choice=None, flow_names=None, column_names=None, metric_group_names=None):
        super(FlowMetricsRequest, self).__init__()
        self._parent = parent
        self._choice = choice
        self._set_property('flow_names', flow_names)
        self._set_property('column_names', column_names)
        self._set_property('metric_group_names', metric_group_names)
```

snappi_ixnetwork/ixnetworkapi.py
alakendu: current code may not handle that as it return after each if

Anyway current choice actually support Union

According to the model
```pyhon
    @property
    def choice(self):
        # type: () -> Union[port, flow, bgpv4, choice, choice, choice]
        """choice getter

        TBD

        Returns: Union[port, flow, bgpv4, choice, choice, choice]
        """
        return self._get_property('choice')
```

ashutshkumr (approved):

PR #254, opened 2021-02-05
snappi unit test map

Rangababu-R:
snappi unit test map

### DIFF
--- a/tests/ipv4/test_ip_priority_field.py
+++ b/tests/ipv4/test_ip_priority_field.py
@@ -1,7 +1,3 @@
-import pytest
-
-
-@pytest.mark.skip("skip until migrated to snappi")
 def test_counter_ip_dscp(api, b2b_raw_config, utils):
     """
     Configure a raw IPv4 flow with,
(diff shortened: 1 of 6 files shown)


tests/ipv4/test_ip_priority_field.py
```diff
@@ -12,26 +8,19 @@ def test_counter_ip_dscp(api, b2b_raw_config, utils):
...
     ]
-
     for i, p in enumerate(phb):
-        dscp = flow_ipv4.Dscp(
-            phb=flow.Pattern(getattr(flow_ipv4.Dscp, p))
-        )
-        prio = flow_ipv4.Priority(dscp)
-        f.packet[1].ipv4.priority = prio
-        utils.apply_config(api, b2b_raw_config)
+        ipv4.priority.choice = 'dscp'
```
ashutshkumr, line 21: we don't need to set a choice here.

tests/ipv4/test_ip_priority_field.py
```diff
@@ -55,11 +43,13 @@ def test_ip_priority_tos(api, b2b_raw_config, utils):
     """
 
     f = b2b_raw_config.flows[0]
-    f.packet = [
-        flow.Header(flow.Ethernet()),
-        flow.Header(flow.Ipv4())
-    ]
-    utils.apply_config(api, b2b_raw_config)
+    # import snappi
```
ashutshkumr, line 46: these comments are not needed anymore i suppose, so remove them

ashutshkumr (changes requested):
please address minor changes

tests/ipv4/test_ip_priority_field.py
Rangababu-R: directly setting phb value throwing attribute error
>>> print(i)
{}
>>> i.priority.dhcp.phb.value = 10
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'FlowIpv4Priority' object has no attribute 'dhcp'
>>>

tests/ipv4/test_ip_priority_field.py
Rangababu-R: yes missed to remove it.

tests/ipv4/test_ip_priority_field.py
ashutshkumr: it should be `dscp`.

ashutshkumr (approved):

PR #257, opened 2021-02-05
Misc fixes and restructure 

alakendu:


snappi_ixnetwork/ngpf.py
```diff
@@ -80,6 +87,11 @@ def _config_proto_stack(self, ixn_obj, snappi_obj, ixn_dg):
             stack_class = getattr(self, '_configure_{0}'
```
Rangababu-R, line 87: add type check for container_name and name, then raise exception

alakendu: According to plan must filed should validate in Model layer. Anyway I have added check for container_name.

alakendu:
This should address https://github.com/open-traffic-generator/ixnetwork/issues/260

snappi_ixnetwork/ixnetworkapi.py
```diff
@@ -74,9 +74,15 @@ def snappi_config(self):
         return self._config
 
     def get_config_object(self, name):
+        if self._config_objects.get(name):
+            raise NameError("{0} is not avialable any Device object"
+                            .format(name))
         return self._config_objects[name]
```
ashutshkumr, line 81: The condition should be checking for None. But either way, I would propose a different approach.
It's more pythonic to use try/catch here (to avoid accessing dict twice in positive scenario):
```python
try:
    return dict[key]
except KeyError:
    raise NameError(msg)
```

snappi_ixnetwork/ixnetworkapi.py
```diff
@@ -74,9 +74,15 @@ def snappi_config(self):
...
 
     def get_device_encap(self, name):
+        if self._device_encap.get(name) is None:
+            raise NameError("{0} is not avialable any Device object"
```
ashutshkumr, line 84: The error message should be `"snappi object named {0} not found".format(name)`. Feel free to rephrase it, but we should use similar error message in get_config_object and any other func that looks for obj based on keys.

snappi_ixnetwork/ixnetworkapi.py
```diff
@@ -136,16 +142,22 @@ def _request_detail(self):
     def set_config(self, config):
         """Set or update the configuration
         """
+        if isinstance(config, (type(self.config()),
```
ashutshkumr, line 145: `self.config()` is costly, so save it in a var so it could be used during deserialization.

snappi_ixnetwork/ixnetworkapi.py
```diff
@@ -154,31 +166,47 @@ def set_config(self, config):
...
         return self._request_detail()
 
-    def set_transmit_state(self, flow_transmit_state):
+    def set_transmit_state(self, payload):
         """Set the transmit state of flows
         """
-        if isinstance(flow_transmit_state, str) is True:
-            flow_transmit_state = self.transmit_state().deserialize(
-                                    flow_transmit_state)
+        if isinstance(payload, (type(self.transmit_state()),
```
ashutshkumr, line 175: store transmit_state() in var here as well

snappi_ixnetwork/ixnetworkapi.py
```diff
@@ -154,31 +166,47 @@ def set_config(self, config):
...
+        if isinstance(payload, str) is True:
+            payload = self.transmit_state().deserialize(
+                                    payload)
         self._connect()
-        self.traffic_item.transmit(flow_transmit_state)
+        self.traffic_item.transmit(payload)
         return self._request_detail()
 
     def set_link_state(self, link_state):
+        if isinstance(link_state, (type(self.link_state()),
```
ashutshkumr, line 187: store self.link_state() in var here as well

snappi_ixnetwork/ixnetworkapi.py
```diff
@@ -154,31 +166,47 @@ def set_config(self, config):
...
         self._connect()
         if link_state.port_names is not None:
             self.vport.set_link_state(link_state)
         return self._request_detail()
     
-    def set_capture_state(self, request):
+    def set_capture_state(self, payload):
         """Starts capture on all ports that have capture enabled.
         """
+        if isinstance(payload, (type(self.capture_state()),
```
ashutshkumr, line 201: store self.capture_state() in a var

snappi_ixnetwork/ixnetworkapi.py
```diff
@@ -281,6 +309,22 @@ def _backup_errors(self):
         if len(app_errors) > 0:
             self._ixn_errors = app_errors[0].Error.find()
 
+    def _validate_instance(self, config):
+        if self._traffic.State == 'started':
+            msg = "set_config may encounter issue " \
```
ashutshkumr, line 314: please specify exactly what issue users will see

ashutshkumr (changes requested):
After addressing reviews, please sync with latest from snappi_ixnetwork branch (since I see a merge conflict)

snappi_ixnetwork/ixnetworkapi.py
alakendu: change it accordingly

snappi_ixnetwork/ixnetworkapi.py
alakendu: modify error message

snappi_ixnetwork/ixnetworkapi.py
alakendu: Added as variable

snappi_ixnetwork/ixnetworkapi.py
alakendu: added

snappi_ixnetwork/ixnetworkapi.py
alakendu: Added

snappi_ixnetwork/ixnetworkapi.py
alakendu:
```python
            msg = "Flows are in running state. " \
                  "Please stop those using set_transmit_state"
```
please let me know

PR #262, opened 2021-02-08
updating snappi unit tests

ANISH-GOTTAPU:
updating snappi unit tests


ashutshkumr (approved):
We may need to add proper validation for flow sizes and devices (e.g. fetch config from ixn and validate it's as expected)

PR #275, opened 2021-02-10
BGP Metrics implementation

Rangababu-R:
Implemented bgp metrics.
Example:
    req = api.metrics_request()
    req.choice = req.BGPV4
    req.bgpv4.column_names = ['sessions_total', 'sessions_up']
    results = api.get_metrics(req)
    assert len(results.bgpv4_metrics) > 0

### DIFF
--- a/snappi_ixnetwork/protocolmetrics.py
+++ b/snappi_ixnetwork/protocolmetrics.py
@@ -1,0 +1,123 @@
+from snappi_ixnetwork.timer import Timer
+
+
+class ProtocolMetrics(object):
+    # Currently paging of the statistic view is not handled
+    # TODO Need to enhance when device groups statistics reach
+    # more than one page.
+
+    _SUPPORTED_PROTOCOLS_ = [
+        "bgpv4"
+    ]
+    _PROTO_NAME_MAP_ = {
+        "bgpv4": 'BGP Peer Per Port',
+        "bgpv4_drill": 'BGP Peer Drill Down'
+    }
+
+    _RESULT_COLUMNS = {
+        "bgpv4": [
+            ("name", "Device Group", str),
+            ("sessions_total", "Sessions Total", int),
+            ("sessions_up", "Sessions Up", int),
+            ("sessions_down", "Sessions Down", int),
+            ("sessions_not_started", "Sessions Not Started", int),
+            ("routes_advertised", "Routes Advertised", int),
+            ("routes_withdrawn", "Routes Withdrawn", int),
+        ]
+    }
+
+    def __init__(self, ixnetworkapi):
+        super().__init__()
+        self._api = ixnetworkapi
+        self.columns = []
+        self.device_names = []
+
+    def get_supported_protocols(self):
+        """
+        Return the protocols that are supported currently
+        """
+        return self._SUPPORTED_PROTOCOLS_
+
+    def _get_per_port_stat_view(self, protocol):
+        protocol_name = self._PROTO_NAME_MAP_.get(protocol)
+        if protocol_name is None:
+            raise NotImplementedError(
+                "{} is Not Implemented".format(protocol)
+            )
+        try:
+            table = self._api.assistant.StatViewAssistant(protocol_name, 10)
+        except Exception:
+            msg = "Could not retrieve stats view for {}\
+                make sure the protocol is up and running".format(protocol)
+            raise Exception(msg)
+        drill_option = [
+            opt for opt in table.DrillDownOptions()
+            if 'Device Group' in opt
+        ]
+        if len(drill_option) == 0:
+            raise LookupError("Could not fetch drill down option")
+        table._View.DrillDown.find().TargetDrillDownOption = drill_option[0]
+        return table
+
+    def _get_per_device_group_stats(self, protocol):
+        table = self._get_per_port_stat_view(protocol)
+        v = table._View
+        column_names = self._RESULT_COLUMNS.get(protocol)
+        row_lst = list()
+        for row in range(len(table.Rows)):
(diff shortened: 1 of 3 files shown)


tests/bgp/test_bgpv4_stats.py
```diff
@@ -0,0 +1,71 @@
...
+
+    bgp1.dut_ipv4_address.increment.start = '10.1.1.2'
+    bgp1.dut_ipv4_address.increment.step = '0.0.1.0'
+
+    bgp2.dut_ipv4_address.increment.start = '10.1.1.1'
+    bgp2.dut_ipv4_address.increment.step = '0.0.1.0'
+
+    utils.start_traffic(api, b2b_raw_config)
+    req = api.metrics_request()
+    req.choice = req.BGPV4
```
ashutshkumr, line 32: don't use choice here. just do
```python
req.bgpv4.device_names = []
```

tests/bgp/test_bgpv4_stats.py
```diff
@@ -0,0 +1,71 @@
...
+        'rx_bgp': [10, 10, 0, 0, 0, 0]
+    }
+    assert len(results.bgpv4_metrics) == 2
+    for bgp_res in results.bgpv4_metrics:
+        for i, enum in enumerate(enums):
+            val = expected_results[bgp_res.name][i]
+            assert getattr(bgp_res, enum) == val
+
+    req = api.metrics_request()
+    req.choice = req.BGPV4
```
ashutshkumr, line 51: choice shouldn't be needed here

tests/bgp/test_bgpv4_stats.py
```diff
@@ -0,0 +1,71 @@
...
+
+    assert len(results.bgpv4_metrics) == 1
+    assert results.bgpv4_metrics[0].name == 'rx_bgp'
+    for bgp_res in results.bgpv4_metrics:
+        for i, enum in enumerate(enums):
+            val = expected_results[bgp_res.name][i]
+            assert getattr(bgp_res, enum) == val
+
+    req = api.metrics_request()
+    req.choice = req.BGPV4
```
ashutshkumr, line 63: choice shouldn't be needed here

snappi_ixnetwork/protocolmetrics.py
```diff
@@ -0,0 +1,122 @@
...
+        return self._SUPPORTED_PROTOCOLS_
+
+    def _get_per_port_stat_view(self, protocol):
+        protocol_name = self._PROTO_NAME_MAP_.get(protocol)
+        if protocol_name is None:
+            raise NotImplementedError(
+                "{} is Not Implemented".format(protocol)
+            )
+        try:
+            table = self._api.assistant.StatViewAssistant(protocol_name, 10)
```
ashutshkumr, line 47: let's not hardcode the timeout. let's call it metrics_timeout or something and use that variable instead. Also, for now we can use same timeout variable for all kinds of metrics.

snappi_ixnetwork/protocolmetrics.py
```diff
@@ -0,0 +1,122 @@
...
+    # Currently paging of the statistic view is not handled
+    # TODO Need to enhance when device groups statistics reach
+    # more than one page.
+
+    _SUPPORTED_PROTOCOLS_ = [
+        "bgpv4"
+    ]
+    _PROTO_NAME_MAP_ = {
+        "bgpv4": 'BGP Peer Per Port',
+        "bgpv4_drill": 'BGP Peer Drill Down'
```
ashutshkumr, line 14: please add a TODO here mentioning that the fate of this is not yet decided (just as a reminder for us)

snappi_ixnetwork/protocolmetrics.py
ashutshkumr: BTW, don't keep the default timeout to 10. Maybe 30.
We can reduce it later, but for now I'd like to be conservative. What do you think ? (you can add a TODO there too)

snappi_ixnetwork/protocolmetrics.py
ashutshkumr: the reason why I'm saying this is because we may want a data structure like this:

```python
_PROTO_NAME_MAP_ = {
    "bgpv4": {'per_port': '', 'drill_down': ''}
}
```

snappi_ixnetwork/protocolmetrics.py
```diff
@@ -0,0 +1,122 @@
...
+                drill = self._api.assistant.StatViewAssistant(
+                    self._PROTO_NAME_MAP_.get(protocol + '_drill')
+                )
+            except Exception as e:
+                msg = 'Could not retrive drill down view at row index %s'\
+                    % (row)
+                raise Exception(msg + e)
+            for dev_row in drill.Rows:
+                dev_name = dev_row['Device Group']
+                if len(self.names) > 0 and dev_name not in self.names:
```
alakendu, line 79: Is it not possible to check dev_name present in that port. Because StatViewAssistant is really costly operation.

snappi_ixnetwork/protocolmetrics.py
```diff
@@ -0,0 +1,122 @@
...
+                    continue
+                row_dt = dict()
+                for sn, ixn, typ in column_names:
+                    try:
+                        value = dev_row[ixn]
+                    except Exception:
```
alakendu, line 85: We normally set some default value according to data type . Yes it is debatable what is the default value (0 or -1 etc). But I feel string will break customer validation logic.

Rangababu-R:
resolved the comments in other PR #276

alakendu (approved):

Issue #283, opened 2021-02-15
UHD speed support


alakendu:

ashutshkumr:
fixed as per `snappi[ixnetwork]==0.3.13`

PR #284, opened 2021-02-15
UHD speed support

alakendu:
This should address https://github.com/open-traffic-generator/ixnetwork/issues/283

### DIFF
--- a/snappi_ixnetwork/ixnetworkapi.py
+++ b/snappi_ixnetwork/ixnetworkapi.py
@@ -10,6 +10,7 @@ from snappi_ixnetwork.trafficitem import TrafficItem
 from snappi_ixnetwork.capture import Capture
 from snappi_ixnetwork.timer import Timer
 from snappi_ixnetwork.protocolmetrics import ProtocolMetrics
+from snappi_ixnetwork.resourcegroup import ResourceGroup
 
 
 class Api(snappi.Api):
@@ -62,6 +63,7 @@ class Api(snappi.Api):
         self.traffic_item = TrafficItem(self)
         self.capture = Capture(self)
         self.protocol_metrics = ProtocolMetrics(self)
+        self.resourec_group = ResourceGroup(self)
 
     def _get_addr_port(self, host):
         items = host.split('/')
--- a/snappi_ixnetwork/ixnetworkapi.py
+++ b/snappi_ixnetwork/ixnetworkapi.py
@@ -63,7 +63,7 @@ class Api(snappi.Api):
         self.traffic_item = TrafficItem(self)
         self.capture = Capture(self)
         self.protocol_metrics = ProtocolMetrics(self)
-        self.resourec_group = ResourceGroup(self)
+        self.resource_group = ResourceGroup(self)
 
     def _get_addr_port(self, host):
         items = host.split('/')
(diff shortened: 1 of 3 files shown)


snappi_ixnetwork/resourcegroup.py
```diff
@@ -0,0 +1,202 @@
...
+        self.layer1_check = []
+        self._layer1_conf = self._api.snappi_config.layer1
+        if self._layer1_conf  is None or len(
+                self._layer1_conf ) == 0:
+            return
+        
+        response = None
+        try:
+            payload = {"arg1": "/availableHardware", "arg2" : []}
+            url = '%s/availableHardware/operations/getChassisWithDetailedResouceGroupsInfo' \
```
Rangababu-R, line 33: does this work for other than UHD as well? ex: for windows and linux API servers

snappi_ixnetwork/vport.py
```diff
@@ -308,15 +312,26 @@ def _set_location(self):
         # calling little bit costly operation? Otherwise we can't handle same port config for
         # multiple run (_set_card_resource_mode reset to card level). Also some ports are
         # handling multiple speed ('novusFourByTwentyFiveGigNonFanOut' and 'novusFourByTenGigNonFanOut')
-        imports = []
+
+        # todo: restructure old module to getChassisWithDetailedResouceGroupsInfo
+        is_uhd = False
+        for port in self._api.snappi_config.ports:
+            location = port.location
+            if '/' in location:
```
Rangababu-R: use chassisobj.chassisType from restpy instead of if '/' from location to set is_uhd. getting from chassis is more accurate than from location format passed

snappi_ixnetwork/ixnetworkapi.py
```diff
@@ -62,6 +63,7 @@ def __init__(self,
         self.traffic_item = TrafficItem(self)
         self.capture = Capture(self)
         self.protocol_metrics = ProtocolMetrics(self)
+        self.resourec_group = ResourceGroup(self)
```
Rangababu-R, line 66: resource spell check self.resourec

Rangababu-R (changes requested):

snappi_ixnetwork/resourcegroup.py
alakendu: According to my understanding UHD is appliance base. This is is expose for chassis base as well. But current snappi_ixnetwork code is use this path only for UHD. I have a plan restructure those. This PR only for UHD.

snappi_ixnetwork/vport.py
alakendu: I also think that as well. But we have several other module also follow same UHD notation say Star. Some of them also applicable in chassis as well. So port may come for them. Then customer may use old and new format card in same test.
Anyway, please let me know consider those use case. I will update accordingly.

snappi_ixnetwork/ixnetworkapi.py
alakendu: Thanks!

snappi_ixnetwork/vport.py
Rangababu-R: In that case we will analyze more on the notations and come up with the common approach later.
for now we will stick to your approach.

alakendu: Thanks!

ashutshkumr:
Approved by tests done be @Rangababu-R

PR #289, opened 2021-02-18
restructure speed aggregation

alakendu:


snappi_ixnetwork/resourcegroup.py
```diff
@@ -173,19 +217,34 @@ def port_name(self):
...
+                'appId'].lower()) is not None:
+            map_speed = 'normal'
         else:
-            raise Exception("Speed %s not avialable within internal map" %
-                              speed)
+            try:
+                map_speed = Vport._SPEED_MODE_MAP[
+                        self._speed]
+            except Exception:
+                raise Exception("Speed %s not avialable within internal map" %
```
ajbalogh, line 238: typo - should be 'available'

snappi_ixnetwork/ixnetworkapi.py
```diff
@@ -285,6 +285,29 @@ def add_error(self, error):
         else:
             self._errors.append(error)
 
+    def parse_location_info(self, location):
+        """It will return (chassis,card,port)
+        set card as 0 where that is not applicable"""
+        if re.search("/|;", location) is None:
+            raise Exception("Please specify location with supported connection status")
```
Rangababu-R, line 292: exception message is not clear, as the condition is search for "/|;", message shall be some thing like
"Invalid port location format, expected [chassis_ip;card_id;port_id or chassis_ip/port_id(UHD only)]"

snappi_ixnetwork/ixnetworkapi.py
```diff
@@ -285,6 +285,29 @@ def add_error(self, error):
...
+        LocationInfo = namedtuple("LocationInfo", ["chassis_info",
+                                                   "card_info",
+                                                   "port_info"])
+        if ';' in location:
+            try:
+                (chassis_info, card_info, port_info) = location.split(';')
+            except Exception:
+                raise Exception("Please specify physical <chassis>;<card>;<port>")
```
Rangababu-R, line 300: remove word "physical" as it refers to physical hardware, which would miss lead for vm

snappi_ixnetwork/ixnetworkapi.py
```diff
@@ -285,6 +285,29 @@ def add_error(self, error):
...
+        else:
+            try:
+                card_info = 0
+                (chassis_info, port_info) = location.split('/')
+            except Exception:
+                raise Exception("Please specify physical <chassis>/<port>")
```
Rangababu-R, line 306: remove work "physical" and mention "UHD only"

snappi_ixnetwork/resourcegroup.py
```diff
@@ -36,25 +27,36 @@ def set_group(self):
...
-        self._process_property()
-        try:
-            chassis_id = 1
-            result = response['result'][0]
-            self._process_groups(result['cards'], chassis_id)
-        except Exception:
-            raise Exception("Problem to parse chassis details")
+        chassis_id = 1
+        for result in response['result']:
+            chassis_dns = result["dns"]
```
Rangababu-R, line 32: does key "dns" always present in the result? if key "dns" is not available it will throw key error.
avoid it by result.get("dns")

Rangababu-R (changes requested):

snappi_ixnetwork/resourcegroup.py
alakendu: I found "dns" present most of our chassis and appliance. And as this is our internal IxNetwok call probably that should applicable for other type. Anyway I have global exception block which will handle these kind of excption.

Rangababu-R (approved):

Issue #280, opened 2021-02-11
Do not allow certain version of IxNetwork to work with snappi_ixnetwork


ashutshkumr:

ashutshkumr:
snappi shouldn't work with ixnetwork older than 9.10

PR #333, opened 2021-04-06
ixn version check

alakendu:
This should address https://github.com/open-traffic-generator/ixnetwork/issues/280

### DIFF
--- a/snappi_ixnetwork/ixnetworkapi.py
+++ b/snappi_ixnetwork/ixnetworkapi.py
@@ -368,7 +368,7 @@ class Api(snappi.Api):
         major, minor = self._globals.BuildNumber.split('.')[0:2]
         if int(major) < 9:
             return False
-        if int(major) >= 9 and int(minor) < 10:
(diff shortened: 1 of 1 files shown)


snappi_ixnetwork/ixnetworkapi.py
```diff
@@ -362,6 +364,14 @@ def _connect(self):
                 self.warning('{}'.format(e))
         self._backup_errors()
 
+    def _ixn_version_check(self):
+        major, minor = self._globals.BuildNumber.split('.')[0:2]
+        if int(major) < 9:
+            return False
+        if int(major) >= 9 and int(minor) < 10:
```
ashutshkumr, line 371: Won't this raise error for version `10.8` ?

ashutshkumr (approved):

snappi_ixnetwork/ixnetworkapi.py
alakendu: int(major) == 9 and int(minor) < 10

Issue #354, opened 2021-04-20
choice support for flow.duration


alakendu:
Please check this PR for details
https://github.com/open-traffic-generator/models/pull/105

alakendu:
Issue should not present in v0.4.11

PR #355, opened 2021-04-20
Choice delay gap

alakendu:
This should address https://github.com/open-traffic-generator/ixnetwork/issues/354


ashutshkumr (approved):

ashutshkumr (changes requested):
we shouldn't round gap values to int

```python
args['StartDelay'] = int(value) * 1000
args['InterBurstGap'] = int(value) * 1000
```

ashutshkumr (approved):

Issue #356, opened 2021-04-22
Experiencing an intermittent error when running a sonic test


alakendu:
![exception](https://user-images.githubusercontent.com/35574510/115694799-d46c3480-a37e-11eb-9b42-8ab85c888371.png)

[Andy]
I noticed that the stack trace executes PauseStatelessTraffic.
I'm able to reproduce this using PauseStatelessTraffic using the rest api.
I start traffic on a configuration then stop it.
I then use a rest client to execute the pauseStatelessTraffic and it gives me that exact error.

ANISH-GOTTAPU:
issue verified with `ixnetwork-open-traffic-generator==0.0.79`

PR #357, opened 2021-04-22
Handle PauseStatelessTraffic for stopped state

alakendu:
This PR probably handle this issue https://github.com/open-traffic-generator/ixnetwork/issues/356

As we are not able to reproduced this issue in house using OTG script. But we are suspecting some flow probably have very few duration. Therefore those flow are stop instantaneously. And we are trying to start using PauseStatelessTraffic(False). 

As a solution we move PauseStatelessTraffic(False) condition. As well as put that within exception block. We will raise an warning and stop those flows when we are getting this type of error.

Code have StartStatelessTrafficBlocking() which will again start all those flows.

### DIFF
--- a/ixnetwork_open_traffic_generator/trafficitem.py
+++ b/ixnetwork_open_traffic_generator/trafficitem.py
@@ -554,14 +554,11 @@ class TrafficItem(CustomField):
         self._api._traffic_item.find(Name=regex)
         if len(self._api._traffic_item) > 0:
             if request.state == 'start':
-                self._api._traffic_item.find(Name=regex, State='^started$')
+                self._api._traffic_item.find(Name=regex, Suspend=True,
+                                             State='^started$')
                 if len(self._api._traffic_item) > 0:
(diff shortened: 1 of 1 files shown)


ixnetwork_open_traffic_generator/trafficitem.py
```diff
@@ -551,14 +554,18 @@ def transmit(self, request):
         self._api._traffic_item.find(Name=regex)
         if len(self._api._traffic_item) > 0:
             if request.state == 'start':
+                self._api._traffic_item.find(Name=regex, State='^started$')
```
ajbalogh, line 557: Would State='^started$', Suspend=True find all items that paused?
Then there shouldn't be a need for a try?

alakendu: Thanks!

ajbalogh (commented):
lgtm

ashutshkumr (approved):

PR #377, opened 2021-06-02
Auto Negotation is not applied properly with Python 2.7

alakendu:


ixnetwork_open_traffic_generator/vport.py
```diff
@@ -620,11 +620,17 @@ def _set_gigabit_auto_negotiation(self, vport, layer1, imports):
...
         if re.search('novustengiglan', vport['type'].lower()) is not None:
             auto_field_name = 'autoNegotiate'
+        # Due to ieeeL1Defaults dependency
+        ieee_media_defaults = {
+            'xpath':
+                vport['xpath'] + '/l1Config/' + vport['type'].replace('Fcoe', ''),
+            'ieeeL1Defaults':
+                layer1.ieee_media_defaults
+        }
+        self._add_l1config_import(vport, ieee_media_defaults, imports)
```
ANISH-GOTTAPU, line 630: `_add_l1config_import` doesn't import instantly and just appends to the imports, I think we will end up with same problem.

alakendu: I feel that should work as we pass as a list of dic:
[{'xpath': '/vport[1]/l1Config/novusHundredGigLan', 'ieeeL1Defaults': False}, {'xpath': '/vport[1]/l1Config/novusHundredGigLan', 'enableAutoNegotiation': False, 'enableRsFec': False, 'linkTraining': False}, {'xpath': '/vport[2]/l1Config/novusHundredGigLan', 'ieeeL1Defaults': False}, {'xpath': '/vport[2]/l1Config/novusHundredGigLan', 'enableAutoNegotiation': False, 'enableRsFec': False, 'linkTraining': False}]
Could you please validate. Otherwise I will add import call in between. I do not want to add multiple import which will add more execution time.

ajbalogh: I agree with @alakendu.  As importconfig requires the xpath for each item you have the flexibility of specifying a list of dicts which should resolve the issue of order. The list of dicts should have the ieeedefaults attribute before the others.

ajbalogh (approved):

ANISH-GOTTAPU (approved):

PR #389, opened 2021-06-24
adding convergence UT

ANISH-GOTTAPU:
adding convergence unit test


ashutshkumr:
Already merged as part of snappi_restructure PR

PR #392, opened 2021-06-28
fix auto_negotiation

alakendu:


ANISH-GOTTAPU:
https://github.com/open-traffic-generator/ixnetwork/blob/main/snappi_ixnetwork/snappi_convergence_api.py#L43

This exception on this line is misleading

The exception is raised for self._api.config_ixnetwork(config) in vport, can you check what is the best place to raise exception and catch the error

ashutshkumr (approved):

PR #396, opened 2021-06-30
add pretest

ANISH-GOTTAPU:
add pretest

### DIFF
--- a/do.py
+++ b/do.py
@@ -36,15 +36,6 @@ def lint():
 
 
 def test():
-    pretest_args = [
-        '--location="https://10.39.71.97:443"',
-        (
-            '--ports="10.39.65.230;6;1 10.39.65.230;6;2 10.39.65.230;6;3'
-            ' 10.39.65.230;6;4"'
-        ),
-        '--media="fiber"',
-        "tests/test_pretest.py",
-    ]
     args = [
         '--location="https://10.39.71.97:443"',
         (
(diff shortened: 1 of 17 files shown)


do.py
```diff
@@ -36,6 +36,15 @@ def lint():
 
 
 def test():
+    pretest_args = [
```
ashutshkumr, line 39: we shouldn't write another test and run them separately. 

the goal is to run some code automatically before starting all tests. This can be done using pytest fixtures.

https://github.com/pytest-dev/pytest/issues/3709#issuecomment-406953552

tests/test_pretest.py
```diff
@@ -0,0 +1,25 @@
...
+    2) Remove any previous stale sessions in API server
+
+    """
+    config = api.config()
+
+    for i in range(len(utils.settings.ports)):
+        config.ports.port(name="port%d" % i, location=utils.settings.ports[i])[
+            -1
+        ]
+
```
ashutshkumr, line 18: Have we tested that this works if we cancel pipeline while the test is running and then restart pipeline ?

ashutshkumr (approved):
Use pytest fixtures instead of a separate pytest script to do cleanup before running tests.

ashutshkumr (changes requested):
Use pytest fixtures instead of a separate pytest script to do cleanup before running tests.

ashutshkumr:
Closing as https://github.com/open-traffic-generator/snappi-ixnetwork/pull/399 already fixes the underlying issue.

ANISH-GOTTAPU:
@alakendu & @Rangababu-R : Please suggest any pointer discussed during the meeting

ashutshkumr (approved):

PR #397, opened 2021-07-01
changes in traffic performance

Rangababu-R:


snappi_ixnetwork/trafficitem.py
```diff
@@ -1,4 +1,4 @@
-import time
+import json
 from snappi_ixnetwork.timer import Timer
 from snappi_ixnetwork.customfield import CustomField
```
alakendu, line 3: We need to modify customfield.py

snappi_ixnetwork/trafficitem.py
```diff
@@ -84,31 +80,48 @@ class TrafficItem(CustomField):
...
+        "class_enable_vector": "pfcPause.header.macControl.priorityEnableVector",
+        "pause_class_0": "pfcPause.header.macControl.pauseQuanta.pfcQueue0",
+        "pause_class_1": "pfcPause.header.macControl.pauseQuanta.pfcQueue1",
+        "pause_class_2": "pfcPause.header.macControl.pauseQuanta.pfcQueue2",
+        "pause_class_3": "pfcPause.header.macControl.pauseQuanta.pfcQueue3",
+        "pause_class_4": "pfcPause.header.macControl.pauseQuanta.pfcQueue4",
+        "pause_class_5": "pfcPause.header.macControl.pauseQuanta.pfcQueue5",
+        "pause_class_6": "pfcPause.header.macControl.pauseQuanta.pfcQueue6",
+        "pause_class_7": "pfcPause.header.macControl.pauseQuanta.pfcQueue7",
+        "order": [
```
alakendu, line 94: do we have any issue if we are not setting in order

snappi_ixnetwork/trafficitem.py
```diff
@@ -191,9 +220,245 @@ class TrafficItem(CustomField):
...
+        for name in dev_names:
+            href = self._api.get_ixn_href(name)
+            if href is None:
+                # TODO change to more detailed message
+                raise Exception(
+                    "href is not found for device name {}".format(name)
+                )
+            href = href.split("ixnetwork")[-1]
+            url, search = self._get_search_payload(
+                href, "^(xpath|ethernet|ipv4|ipv6|bgpv4|bgpv6)$", ["name"], []
```
alakendu, line 327: I think we need only xpath which select from device href. So is there any specific reason to provide this filter "^(xpath|ethernet|ipv4|ipv6|bgpv4|bgpv6)$"

snappi_ixnetwork/trafficitem.py
```diff
@@ -191,9 +220,245 @@ class TrafficItem(CustomField):
...
+        ixn = self._api.assistant._ixnetwork
+        result = ixn._connection._execute(url, payload)
+        if len(dev_names) != len(result):
+            # TODO change to more detailed message
+            raise Exception("search query result miss match with device names")
+        paths = {}
+        for i, d in enumerate(dev_names):
+            paths[d] = {"xpath": result[i]["xpath"]}
+            string = str(result[i])
+            if "ipv4" in string:
```
alakendu, line 340: We can referred get_device_encap rather than string search. Consider case when IPv6 route is configure on top of IPv4

snappi_ixnetwork/trafficitem.py
```diff
@@ -203,35 +468,55 @@ def config(self):
...
-                    "Name": flow.name,
-                    "TrafficItemType": "l2L3",
-                    "TrafficType": self._get_traffic_type(flow),
-                    "SrcDestMesh": self._get_mesh_type(flow),
-                }
-                ixn_traffic_item.find(
-                    Name="^%s$" % self._api.special_char(flow.name),
-                    TrafficType=args["TrafficType"],
+        self.remove_ixn_traffic()
+        if len(self._api._config.flows) == 0:
```
alakendu, line 472: better use public self._api.snappi_config

snappi_ixnetwork/trafficitem.py
```diff
@@ -191,9 +220,245 @@ class TrafficItem(CustomField):
...
+                    "for flow indexed at %s" % i
+                )
+                raise Exception(msg)
+            if f.tx_rx.choice is None:
+                msg = "Flow endpoint needs to be either port or device"
+                raise Exception(msg)
+            if f.tx_rx.choice == "port":
+                tr_type = "raw"
+                ep = getattr(f.tx_rx, "port")
+                tx_objs = ["%s/protocols" % ports.get(ep.tx_name)]
```
alakendu, line 414: "SrcDestMesh" section is missing which is handle by self._get_mesh_type(flow)

snappi_ixnetwork/timer.py
```diff
@@ -1,5 +1,7 @@
 import time
 
+timer_data = {}
+
 
 class Timer:
```
alakendu, line 6: as a lite note please do not use single char as variable name when that will use in multiple line.

snappi_ixnetwork/trafficitem.py
```diff
@@ -302,277 +578,133 @@ def _configure_options(self):
...
+            ind = stack_names.index(header._choice)
+            ixn_fields = ixn_stack[ind]["field"]
+            fields = self._configure_stack_fields(ixn_fields, header)
+            stacks[ind]["field"] = fields
+        return stacks
+
+    def _append_header(self, snappi_header, xpath, stacks, add_trailer=True):
+        field_map = getattr(self, "_%s" % snappi_header._choice.upper())
+        stack_name = self._HEADER_TO_TYPE.get(snappi_header._choice)
+        if stack_name is None:
```
alakendu, line 611: Not setting default field (packet.get(packet_field_name, with_default=True)). 

Please check performance after adding this one.

snappi_ixnetwork/trafficitem.py
```diff
@@ -191,9 +220,245 @@ class TrafficItem(CustomField):
...
+                rx_objs = [devices[n]["xpath"] for n in ep.rx_names]
+            tr_xpath = "/traffic/trafficItem[%d]" % self.traffic_index
+            tr["trafficItem"].append(
+                {"xpath": tr_xpath, "name": "%s" % f.name}
+            )
+
+            tr["trafficItem"][-1]["endpointSet"] = [
+                {"xpath": tr["trafficItem"][-1]["xpath"] + "/endpointSet[1]"}
+            ]
+            tr["trafficItem"][-1]["endpointSet"][0]["sources"] = [
```
alakendu, line 429: same device within Tx and Rx validation is missing. Please referred _endpoint_validation

snappi_ixnetwork/trafficitem.py
```diff
@@ -191,9 +220,245 @@ class TrafficItem(CustomField):
...
+            # "arg5": True,
+        }
+        try:
+            # TODO for larger config rest api is throwing error,
+            # with no url found, when the first response is 202 (in-progress)
+            # its keep checking the status of the url with 1 sec sleep, and
+            # after a while error is thrown. but could see the configuration
+            # applied at Ixnetwork. (Need to check with Eng team)
+            self._api._request("POST", url=url, payload=payload)
+        except Exception:
```
alakendu, line 282: I am not sure how we will handle as json import error are very generic. User also may not map IxNet internal object with sanppi object. Anyway, first put all error. Then we will decide.

snappi_ixnetwork/trafficitem.py
```diff
@@ -191,9 +220,245 @@ class TrafficItem(CustomField):
...
+            paths[d]["type"] = tr_type
+        return paths
+
+    def get_ixn_config(self, config):
+        ixn = self._api.assistant._ixnetwork
+        myfilter = [{"property": "name", "regex": ".*"}]
+        url, payload = self._get_search_payload(
+            "/traffic",
+            "(?i)^(trafficItem|configElement|frameRate"
+            "|frameSize|transmissionControl|stack|field|highLevelStream"
```
alakendu, line 355: probably stack|field are not require. As those are adding more payload. We will ignore those.

snappi_ixnetwork/trafficitem.py
```diff
@@ -203,35 +468,55 @@ def config(self):
...
+            self.flows_has_loss = []
+            self.latency_mode = None
+            if ixn_traffic_item.get("trafficItem") is None:
+                # TODO raise Exception
+                return
+            ixn_traffic_item = ixn_traffic_item.get("trafficItem")
+            tr_json = {"traffic": {"xpath": "/traffic", "trafficItem": []}}
+            # with Timer(self._api, "json processing for traffic"):
+            for i, flow in enumerate(self._api._config.flows):
+                tr_item = {"xpath": ixn_traffic_item[i]["xpath"]}
```
alakendu, line 487: Traffic Items are accessing according to the order. So access through index ("i") probably do not have any issue. I feel it will better to access that using "name" rather than index ("i")

snappi_ixnetwork/trafficitem.py
```diff
@@ -191,9 +220,245 @@ class TrafficItem(CustomField):
...
+        config_elem = {"xpath": ce_path, "stack": []}
+        for i, header in enumerate(packet):
+            stack_name = self._HEADER_TO_TYPE.get(header._choice)
+            header_xpath = "%s/stack[@alias = '%s-%d']" % (
+                ce_path,
+                stack_name,
+                i + 1,
+            )
+            self._append_header(header, header_xpath, config_elem["stack"])
+        fcs_xpath = "%s/stack[@alias = '%s-%d']" % (
```
alakendu, line 454: This is not valid if stack do not have any Ethernet stack.(say PfcPause)

snappi_ixnetwork/trafficitem.py
Rangababu-R: Yes, right now custom field is not used, while implementing ethernet pause and ipv4 priority, will modify

snappi_ixnetwork/trafficitem.py
Rangababu-R: yes, if we are not following the order, the values are not getting reflected.

snappi_ixnetwork/trafficitem.py
Rangababu-R: This is used to determine the traffic end point type, just with device href we cannot determine the traffictype, if we have these in place we can determine if ipv4, ipv6 and ethernetvlan.
But currently, this pattern is executed for all the hrefs. this can be modified to more specific. will update the code accordingly.
Thanks for mentioning

snappi_ixnetwork/trafficitem.py
Rangababu-R: sure will change the name

snappi_ixnetwork/trafficitem.py
Rangababu-R: agree

snappi_ixnetwork/trafficitem.py
Rangababu-R: yes mesh handling is not added, because this implementation was way older when we are defaulting the mesh type to manytomany. will add mesh handling

snappi_ixnetwork/timer.py
Rangababu-R: I did not get his Alakendu, can you give some example

snappi_ixnetwork/trafficitem.py
Rangababu-R: self._HEADER_TO_TYPE.get ==> this is different "get" method which gets the _HEADER_TO_TYPE dict.
I assume you are pointing at the default header if user doesn't create any header then we shall default it to ethernet?

snappi_ixnetwork/trafficitem.py
Rangababu-R: yes, this validation needs to add as this code is was way older. I missed it. Thanks for pointing.

snappi_ixnetwork/trafficitem.py
Rangababu-R: this was time being addition to make sure all the tests passes. once that is done. I will add the check if the exception raised after the 201[Inprogress] response. which needs a exception handling in snappi_api.py _requests method

snappi_ixnetwork/trafficitem.py
Rangababu-R: traffic.py happens in 2 steps.
1. create the flows.
2. update the parameters of flows (exp: size, rate, duration, stack fields)
this will give the structure, will update the key value pairs.

snappi_ixnetwork/trafficitem.py
Rangababu-R: as this approach is, remove traffic then create. getting the flows with name will add one more code block like validating the name. which would be a over head

snappi_ixnetwork/trafficitem.py
Rangababu-R: Agree. need to handle for pfcpause only. for ethernet or ethenetpause fcs will be same.
I think Ixnetwork has only these 2 l2 headers. correct me if I am wrong

snappi_ixnetwork/trafficitem.py
```diff
@@ -191,9 +220,245 @@ class TrafficItem(CustomField):
...
+                {"xpath": tr["trafficItem"][-1]["xpath"] + "/endpointSet[1]"}
+            ]
+            tr["trafficItem"][-1]["endpointSet"][0]["sources"] = [
+                o for o in tx_objs
+            ]
+            tr["trafficItem"][-1]["endpointSet"][0]["destinations"] = [
+                o for o in rx_objs
+            ]
+            tr["trafficItem"][-1]["trafficType"] = tr_type
+            if tr_type == "raw":
```
alakendu, line 436: In this new implementation we will add additional stack for device endpoint. 
eth, vlan, tcp = flow.packet.ethernet().vlan().tcp()
Then it will produced eth> IPv4 > vlan > tcp (device configure with IPv4)
This actually regress current implementation

Rangababu-R: in the latest implementation, this is handled. but still we can validate once

Rangababu-R:
having merge conflicts so closing this branch and creating new one,

PR #401, opened 2021-07-06
Add remote_link_failure

alakendu:


snappi_ixnetwork/snappi_convergence_api.py
```diff
@@ -97,6 +103,10 @@ def set_state(self, payload):
             self._api._connect()
             if payload.choice is None:
                 raise Exception("state [transmit/ link/ route] must configure")
+            if self._remote_link_failure is True:
```
ashutshkumr, line 106: Raise exception only when route or link state is set.

snappi_ixnetwork/snappi_convergence_api.py
```diff
@@ -97,6 +102,15 @@ def set_state(self, payload):
             self._api._connect()
             if payload.choice is None:
                 raise Exception("state [transmit/ link/ route] must configure")
+            if (
+                self._cvg_event_name is not None
+                and payload.choice != "transmit"
+            ):
+                if re.search(payload.choice, self._cvg_event_name) is None:
+                    raise Exception(
+                        "set_state should not execute as %s is configured and want to performed %s"
```
ANISH-GOTTAPU, line 111: E           snappi_ixnetwork.exceptions.SnappiIxnException:   File "c:\anish\snappi-ixnetwork\snappi_ixnetwork\snappi_convergence_api.py", line 110, in set_state
E               raise Exception(
E            set_state should not execute as link_down is configured and want to performed route

can you change the exception to more readable format and `payload.choice` returns as `route` or `link` instead of `route_withdraw` or `link_down`.

alakendu: added

snappi_ixnetwork/snappi_convergence_api.py
alakendu: Add check

PR #403, opened 2021-07-06
Traffic performance enhancement with json

Rangababu-R:
closed other PR as it has merge conflicts, this PR has all comments addressed raised by Alakendu

### DIFF
--- a/snappi_ixnetwork/trafficitem.py
+++ b/snappi_ixnetwork/trafficitem.py
@@ -146,7 +146,7 @@ class TrafficItem(CustomField):
         "cfi": "vlan.header.vlanTag.cfi",
         "priority": "vlan.header.vlanTag.vlanUserPriority",
         "protocol": "vlan.header.protocolID",
-        "order": ["id", "cfi", "priority", "protocol"],
+        "order": ["priority", "cfi", "id", "protocol"],
     }
 
     _IPV4 = {
@@ -360,6 +360,7 @@ class TrafficItem(CustomField):
             response["result"].get("errata") is not None
             and response["result"]["errata"] != []
         ):
+            self._api.get_json_import_errors()
             raise SnappiIxnException(
                 400, "{}".format(response["result"]["errata"])
             )
@@ -407,6 +408,8 @@ class TrafficItem(CustomField):
             query = "^(xpath|ethernet|ipv4|ipv6)$"
             if "ipv4" in href or "ipv6" in href:
                 query = "^(xpath)$"
+            elif "networkGroup" in href:
+                query = "^(xpath|networkGroup|ipv4PrefixPools|ipv6PrefixPools)"
             url, search = self._get_search_payload(href, query, ["name"], [])
             selects.extend(search["selects"])
         payload = {"selects": selects}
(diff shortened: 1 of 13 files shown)


ANISH-GOTTAPU (commented):
Tests to be added
1) Multiple devices with route range in a single topology
2) Validate traffic flow rate
3) Validate traffic size & duration

snappi_ixnetwork/trafficitem.py
```diff
@@ -182,18 +284,311 @@ class TrafficItem(CustomField):
...
+        for name in dev_names:
+            href = self._api.get_ixn_href(name)
+            if href is None:
+                # TODO change to more detailed message
+                raise Exception(
+                    "href is not found for device name {}".format(name)
+                )
+            href = href.split("ixnetwork")[-1]
+            query = "^(xpath|ethernet|ipv4|ipv6)$"
+            if "ipv4" in href or "ipv6" in href:
```
alakendu, line 413: We already stored that info in self._api.._device_encap. So we will get the type using self._api.get_device_encap.
Otherwise we need to use more regex to extract actual type.

snappi_ixnetwork/trafficitem.py
```diff
@@ -182,18 +284,311 @@ class TrafficItem(CustomField):
...
+            selects.extend(search["selects"])
+        payload = {"selects": selects}
+        ixn = self._api.assistant._ixnetwork
+        result = ixn._connection._execute(url, payload)
+        if len(dev_names) != len(result):
+            # TODO change to more detailed message
+            raise Exception("search query result miss match with device names")
+        paths = {}
+        for i, d in enumerate(dev_names):
+            paths[d] = {"xpath": result[i]["xpath"]}
```
alakendu, line 427: Personally I preferred use self._api.get_device_encap. Because, we are planning to incorporate ISIS as well as unconnected interface. But if you filled searching is ok then I am fine.

snappi_ixnetwork/trafficitem.py
```diff
@@ -203,34 +598,59 @@ def config(self):
...
+                self._configure_size(
+                    tr_item["configElement"], flow.get("size")
                 )
-                if len(ixn_traffic_item) == 0:
-                    ixn_traffic_item.add(**args)
-                else:
-                    self._update(ixn_traffic_item, **args)
-                self._configure_endpoint(
-                    ixn_traffic_item.EndpointSet, flow.tx_rx
+                self._configure_rate(
```
alakendu, line 630: get with_default=True for rate, size and duration. Otherwise it will configure IxNetwork default. And I think we are decide to configure with snappi default value

snappi_ixnetwork/trafficitem.py
Rangababu-R: is the device type stored in current main branch or was it added in device compaction?

snappi_ixnetwork/trafficitem.py
Rangababu-R: I remember adding the defaults. let me check again

snappi_ixnetwork/trafficitem.py
alakendu: This was added log back probably day 1. I copy (self._api.get_device_encap) that from your brunch.

ashutshkumr (approved):

PR #406, opened 2021-07-08
Support IP step in string format

alakendu:
This should address https://github.com/open-traffic-generator/snappi-ixnetwork/issues/405

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,8 +37,8 @@ setuptools.setup(
     install_requires=["ixnetwork-restpy>=1.0.52"],
     extras_require={
         "testing": [
-            "snappi==0.4.15",
-            "snappi_convergence==0.0.18",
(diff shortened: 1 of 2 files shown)


ANISH-GOTTAPU:
There is a relationship between prefix and step

The step is `0.0.0.1` but default prefix is 24. user expectation would be to increment last octet but third octet will increase as default prefix is 24. More discussion is needed on handling this

image

ANISH-GOTTAPU (changes requested):
There is a relationship between prefix and step

The step is `0.0.0.1` but default prefix is 24. user expectation would be to increment last octet but third octet will increase as default prefix is 24. More discussion is needed on handling this

image

alakendu:
We are reverting model and change it to step in <int>

PR #430, opened 2021-08-25
packet header's parent gives none when doing deepcopy

Rangababu-R:


snappi_ixnetwork/trafficitem.py
```diff
@@ -609,6 +617,18 @@ def _endpoint_validation(self, flow):
                     "must be unique for flow %s" % flow.name
                 )
 
+    def copy_flow_packet(self, config):
```
alakendu, line 620: Is it not possible to deepcopy only those packet which is require. And we can do that at the time of creation.
I am ok if there is there is any to do that.

alakendu (approved):

ANISH-GOTTAPU (approved):

PR #446, opened 2021-09-14
Implementation for Ping model

ANISH-GOTTAPU:
Implementation for Ping model


snappi_ixnetwork/ping.py
```diff
@@ -0,0 +1,75 @@
...
+                if req_type == "ipv4":
+                    if src_name not in v4_names:
+                        msg = (
+                            src_name
+                            + """ is not available in the configured v4 interface names """
+                            + str(v4_names)
+                        )
+                        raise Exception(msg)
+                    ip_obj = (
+                        self._api._ixnetwork.Topology.find()
```
alakendu, line 45: It will better to find with some attribute (say Name) to optimize it.

snappi_ixnetwork/ping.py
```diff
@@ -0,0 +1,75 @@
...
+                        .DeviceGroup.find()
+                        .Ethernet.find()
+                        .Ipv6.find(Name=src_name)
+                    )
+                self._api.info("Sending ping to %s" % dst_ip)
+                ping_status = ip_obj.SendPing(DestIP=dst_ip)
+                for reply in ping_status:
```
alakendu, line 66: probably this is a corner case. But check command execution failure due to some exception

alakendu (approved):

snappi_ixnetwork/ping.py
ANISH-GOTTAPU: ok

snappi_ixnetwork/ping.py
ANISH-GOTTAPU: ok

PR #455, opened 2021-11-23
ip protocol field auto to true

ANISH-GOTTAPU:


Rangababu-R:
this fix shall be reverted in 0.6.x

Rangababu-R (approved):

PR #468, opened 2022-01-18
flow metric separation

alakendu:
This should support for port and flow for this issue https://github.com/open-traffic-generator/snappi-ixnetwork/issues/467


ANISH-GOTTAPU:
As discussed, pulling stats using gNMI works if we use ixNetwork windows API Server. The same doesn't work for Linux API Server as the session Id will change.

ANISH-GOTTAPU (approved):

PR #469, opened 2022-01-24
Event Names Reflection for LAG Interfaces

ANISH-GOTTAPU:


ANISH-GOTTAPU:
To Address https://github.com/open-traffic-generator/snappi-ixnetwork/issues/460

alakendu (approved):

PR #474, opened 2022-02-09
support for update flows API on ixnetwork

ANISH-GOTTAPU:
support for update flows API on ixnetwork

### DIFF
--- a/snappi_ixnetwork/snappi_api.py
+++ b/snappi_ixnetwork/snappi_api.py
@@ -16,6 +16,7 @@ from snappi_ixnetwork.exceptions import SnappiIxnException
 from snappi_ixnetwork.device.ngpf import Ngpf
 from snappi_ixnetwork.objectdb import IxNetObjects
 
+
 class Api(snappi.Api):
     """IxNetwork implementation of the abstract-open-traffic-generator package
 
@@ -60,6 +61,7 @@ class Api(snappi.Api):
         self.ixn_objects = None
         self._config_type = self.config()
         self._protocol_state = self.protocol_state()
+        self._flows_update = self.flows_update()
         self._transmit_state = self.transmit_state()
         self._link_state = self.link_state()
         self._capture_state = self.capture_state()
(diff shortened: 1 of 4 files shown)


snappi_ixnetwork/snappi_api.py
```diff
@@ -611,6 +635,9 @@ def _validate_instance(self, config):
                 )
                 self.add_error(msg)
                 self.warning(msg)
+        self._initial_flows_config = self.config().flows.deserialize(
+            config.flows.serialize(config.flows.DICT)
+        )
```
alakendu, line 640: ```serialize``` is a costly operation. It will better if we serialize at the time of update. We already store current config in ```self._running_config```.  So we can utilize it.

snappi_ixnetwork/trafficitem.py
```diff
@@ -1366,3 +1368,97 @@ def _construct_timestamp(self, flow_row, row):
...
+            hl = self._api._ixnetwork.Traffic.TrafficItem.find(
+                Name=flow.name
+            ).HighLevelStream.find()
+            self._update_size(hl, flow.size)
+            self._update_rate(hl, flow.rate)
+
+    def _validate_update_flows_config(self, update_flows_config):
+        errors = []
+        for flow in update_flows_config.flows:
+            if flow not in self._api._config.flows._items:
```
alakendu, line 1387: We may reduced multiple search call by passing all name of traffic items. It will enhance some performance.

snappi_ixnetwork/snappi_api.py
ANISH-GOTTAPU: `self._running_config` is a reference to the config. Since we are updating the same flow object, can't compare the flows to be updated with the existing flows.

We need initial configuration to compare with the existing configuration. Can't think of other thing than serialize, please let me know if there is better approach to handle this.

snappi_ixnetwork/trafficitem.py
ANISH-GOTTAPU: As discussed, we might get different combinations of update rate or size per different flows. Won't be a feasible solution to get all traffic items and apply rate or size per item

alakendu (approved):

PR #485, opened 2022-02-25
support for control capture

ANISH-GOTTAPU:


tests/capture/test_capture_control.py
```diff
@@ -0,0 +1,100 @@
...
+    flow.metrics.enable = True
+
+    utils.start_traffic(api, config)
+
+    request = api.capture_request()
+    request.port_name = "rx"
+    pcap_bytes = api.get_capture(request)
+
+    bgp_pkts = []
+    for _, pkt in dpkt.pcap.Reader(pcap_bytes):
```
alakendu, line 89: I think it will better if we validate some BGP header as well

alakendu (approved):

PR #490, opened 2022-03-02
Support for flow stats in 0.7.x

ANISH-GOTTAPU:
Support for flow stats in 0.7.x

### DIFF
--- a/do.py
+++ b/do.py
@@ -53,8 +53,8 @@ def test():
         (
             '--ports="otg-novus100g.lbj.is.keysight.com;1;1'
             " otg-novus100g.lbj.is.keysight.com;1;2"
-            " otg-novus100g.lbj.is.keysight.com;1;3"
-            ' otg-novus100g.lbj.is.keysight.com;1;4"'
+            " otg-novus100g.lbj.is.keysight.com;1;5"
+            ' otg-novus100g.lbj.is.keysight.com;1;6"'
         ),
         "--ext=ixnetwork",
         "--speed=speed_100_gbps",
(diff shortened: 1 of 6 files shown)


snappi_ixnetwork/trafficitem.py
```diff
@@ -1277,48 +1293,97 @@ def results(self, request):
...
-                flow_row = flow_rows[name]
+        if self._api._flow_tracking:
+            table = self._api.assistant.StatViewAssistant("Flow Statistics")
+            for row in table.Rows:
+                name = row["Traffic Item"]
                 if (
-                    float(row["Tx Frame Rate"]) > 0
-                    or int(row["Tx Frames"]) == 0
+                    len(flow_names) > 0
+                    and row["Traffic Item"] not in flow_names
```
Rangababu-R, line 1302: Why is the logical statement made to "AND" logic instead of "OR"?

Rangababu-R (changes requested):

snappi_ixnetwork/trafficitem.py
ANISH-GOTTAPU: even though flow_names>0 condition satisfies, the `flow_name` requet by the user should be part of flow_names

Rangababu-R (approved):

do.py
```diff
@@ -53,8 +53,8 @@ def test():
         (
             '--ports="otg-novus100g.lbj.is.keysight.com;1;1'
             " otg-novus100g.lbj.is.keysight.com;1;2"
-            " otg-novus100g.lbj.is.keysight.com;1;3"
-            ' otg-novus100g.lbj.is.keysight.com;1;4"'
+            " otg-novus100g.lbj.is.keysight.com;1;5"
```
alakendu, line 56: please revert it when the port will available

ANISH-GOTTAPU: yeah..sure

PR #499, opened 2022-03-10
Trigger Snappi Test after snappi ixnetwork

ASHNA-AGGARWAL-KEYSIGHT:
resolves https://github.com/open-traffic-generator/snappi-ixnetwork/issues/495


alakendu (dismissed):

alakendu:
is it possible to update same snappi version within snappi_test. Otherwise snappi_test will run with older version which is not that useful.

PR #509, opened 2022-03-31
Device chaining Implementation

ASHNA-AGGARWAL-KEYSIGHT:
resolves https://github.com/open-traffic-generator/snappi-ixnetwork/issues/503

### DIFF
--- a/snappi_ixnetwork/device/ngpf.py
+++ b/snappi_ixnetwork/device/ngpf.py
@@ -130,6 +130,8 @@ class Ngpf(Base):
                         print("skip as vxlan_name is not implemented")
                 else:
                     port_name = ethernet.get("port_name")
+                if port_name is None:
+                    raise Exception("port_name is not passed for the device {}".format(device.get("name")))
                 if port_name in self._ixn_topo_objects:
                     ixn_topo = self._ixn_topo_objects[port_name]
                 else:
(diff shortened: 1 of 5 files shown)


snappi_ixnetwork/device/ngpf.py
```diff
@@ -117,9 +117,19 @@ def _configure_topology(self):
...
+                if ethernet.get("connection"):
+                    connection_choice = ethernet.get("connection").choice
+                    if connection_choice == "port_name":
+                        port_name = ethernet.get("connection").port_name
+                    elif connection_choice == "lag_name":
+                        port_name = ethernet.get("connection").lag_name
+                    elif connection_choice == "vxlan_name":
+                        print("skip as vxlan_name is not implemented")
+                else:
+                    port_name = ethernet.get("port_name")
```
ANISH-GOTTAPU, line 132: Throw an error from snappi_ixnetwork library if port_name is not passed at all..Add testcase for the same, so that we can delete the testcase later once `port_name` is deprecated

ASHNA-AGGARWAL-KEYSIGHT: Added the exception handling

ANISH-GOTTAPU (approved):

PR #523, opened 2022-04-20
Issue522: Capture is failing in tests in 9.20

ASHNA-AGGARWAL-KEYSIGHT:
resolves https://github.com/open-traffic-generator/snappi-ixnetwork/issues/522

### DIFF
--- a/snappi_ixnetwork/capture.py
+++ b/snappi_ixnetwork/capture.py
@@ -80,6 +80,8 @@ class Capture(object):
                 }
                 imports.append(capture)
         for capture_item in self._api.snappi_config.captures:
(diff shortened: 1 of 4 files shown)


snappi_ixnetwork/capture.py
```diff
@@ -346,7 +349,7 @@ def results(self, request):
         url = "%s/files?absolute=%s&filename=%s" % (
             self._api._ixnetwork.href,
             path,
-            merged_capture,
+            dc,
```
ANISH-GOTTAPU, line 352: Add a TODO here..revert to merged_capture once fix is available with 9.20

PR #534, opened 2022-05-27
Logger support

alakendu:
resolves https://github.com/open-traffic-generator/snappi-ixnetwork/issues/533


tests/conftest.py
```diff
@@ -1,5 +1,6 @@
 import pytest
 import snappi
+import logging
```
ANISH-GOTTAPU, line 3: N.I.T: Unused import "logging"

ANISH-GOTTAPU (approved):

PR #556, opened 2022-08-23
support for manual gateway mac

### LINKED ISSUE #555: Add Manual Gateway MAC support
Add Manual Gateway MAC support

ANISH-GOTTAPU:
resolves #555

### DIFF
--- a/snappi_ixnetwork/device/base.py
+++ b/snappi_ixnetwork/device/base.py
@@ -122,6 +122,19 @@ class Base(object):
                 ))
             ixn_obj[ixn_attr] = self.multivalue(value)
 
+    def configure_multivalues_with_choice(self, snappi_obj, ixn_obj, attr_map):
+        """attr_map contains snappi_key : ixn_key/ ixn_info in dict format"""
+        self.logger.debug("configuring multivalues:")
+        for snappi_attr, ixn_map in attr_map.items():
+            ixn_attr = ixn_map
+            value = snappi_obj.get(snappi_attr).get("value")
+            if snappi_attr == "gateway_mac":
+                ixn_obj["resolveGateway"] = self.multivalue(False)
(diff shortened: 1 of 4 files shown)


snappi_ixnetwork/device/base.py
```diff
@@ -122,6 +122,19 @@ def configure_multivalues(self, snappi_obj, ixn_obj, attr_map):
                 ))
             ixn_obj[ixn_attr] = self.multivalue(value)
 
+    def configure_multivalues_with_choice(self, snappi_obj, ixn_obj, attr_map):
+        """attr_map contains snappi_key : ixn_key/ ixn_info in dict format"""
+        self.logger.debug("configuring multivalues:")
+        for snappi_attr, ixn_map in attr_map.items():
+            ixn_attr = ixn_map
```
alakendu, line 129: is it require to store `ixn_map` to `ixn_attr`. I am ok if it will require for some other use case

alakendu (approved):

snappi_ixnetwork/device/base.py
```diff
@@ -122,6 +122,19 @@ def configure_multivalues(self, snappi_obj, ixn_obj, attr_map):
                 ))
             ixn_obj[ixn_attr] = self.multivalue(value)
 
+    def configure_multivalues_with_choice(self, snappi_obj, ixn_obj, attr_map):
```
alakendu, line 125: It will better to move `configure_multivalues_with_choice` within interface as this method only work with gateway_mac

ANISH-GOTTAPU: It might require if we get another property with choice in future

snappi_ixnetwork/device/base.py
ANISH-GOTTAPU: yeah..just want this if we get any other property with choice in future

PR #563, opened 2023-05-29
compatible with latest model

desaimg1:
Worked on following issues:
1. Modified warnings as per latest model
2. Added set_control_state and set_control_action as per the latest model
3. Added get_version API


snappi_ixnetwork/protocolmetrics.py
```diff
@@ -387,5 +387,5 @@ def results(self, request):
         if len(self.columns) > 0:
             self.columns.append("name") if "name" not in self.columns else None
             self.columns = list(set(self.columns))
-        with Timer(self._api, "Fetching {} Metrics".format(protocol)):
-            return self._filter_stats(protocol)
+        #with Timer(self._api, "Fetching {} Metrics".format(protocol)):
```
Vibaswan, line 390: any particular reason why timer was removed ?

snappi_ixnetwork/snappi_api.py
```diff
@@ -301,136 +303,94 @@ def _protocols_exists(self):
...
+                    res = self.control_action_response()
+                    self._connect()
+                    res.response.protocol.ipv6.ping.responses.deserialize(self.ping.results(control_choice, request_payload))
+            res.warnings=snappi.Warning()
+            return res
         except Exception as err:
             raise SnappiIxnException(err)
-        return self._request_detail()
+               
+    def set_protocol_state(self, payload):
```
Vibaswan, line 357: This breaks backward compatibility right.
if a old script is using set_protocol_state, now its throws a warning which is right but it wont do any operation right

Vibaswan (changes requested):
I also do not see a place where we have changed the snappi version to the new snappi version we are using, would you please point me to the right location where the change is there ?

snappi_ixnetwork/ping.py
```diff
@@ -19,7 +19,7 @@ class Ping(object):
     def __init__(self, ixnetworkapi):
         self._api = ixnetworkapi
 
-    def results(self, ping_request):
+    def results(self, req_type, ping_request):
```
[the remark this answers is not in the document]
desaimg1: Line number 41

snappi_ixnetwork/ping.py
```diff
@@ -33,9 +33,9 @@ def results(self, ping_request):
                     v6_names.append(ip.name)
 
         with Timer(self._api, "Ping requests completed in"):
-            for endpoint in ping_request.endpoints:
+            for endpoint in ping_request.requests:
                 response = {}
-                req_type = endpoint.parent.choice
+                #req_type = endpoint.parent.choice
```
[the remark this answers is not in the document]
desaimg1: removed

snappi_ixnetwork/protocolmetrics.py
desaimg1: Some of the scripts were failing coz of this. Discussed with Anish and removed

snappi_ixnetwork/snappi_api.py
desaimg1: In new model it is deprecated. Is that we need to think of backward compatibility or the script should be inline with new model? Could you please clarify.

Vibaswan (approved):
Code looks good, Ci failing I guess as ports are not reachable

PR #575, opened 2024-09-27
modified imix script as per the latest model changes

desaimg1:


tests/conftest.py
```diff
@@ -4,10 +4,10 @@
 import utils as utl
 # import snappi_convergence
 
-
-def pytest_exception_interact(node, call, report):
-    # end all pytests on first exception that is encountered
-    pytest.exit(call.excinfo.traceback[0])
+# TODO: uncomment this if we want pipeline to fail instantly
```
Vibaswan, line 7: Please ensure we want to keep this commented code or remove it all together ?

indraniBh (approved):

Vibaswan (approved):

tests/conftest.py
indraniBh: IMO we can remove it. There will be CL in PR, so that we can incorporate if required in future. @desaimg1 @Vibaswan

PR #577, opened 2024-11-13
Blr automation team

desaimg1:
merging ci improvement files

### DIFF
--- a/do.py
+++ b/do.py
@@ -127,9 +127,6 @@ def coverage():
 
     with open("./cov_report/index.html") as fp:
         out = fp.read()
-        result = re.findall(r"data-ratio.*?[>](\d+)\b", out)
-        print(result)
-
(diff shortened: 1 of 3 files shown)


indraniBh:
@desaimg1, Please update with highlights of work done in this PR.

do.py
```diff
@@ -4,8 +4,12 @@
 import sys
 import shutil
 import subprocess
+import smtplib
+from email.mime.multipart import MIMEMultipart
+from email.mime.text import MIMEText
 
 
+global ixnexception
```
alakendu, line 12: is this branch align with main branch? Because I found some changes which is already present in main.

alakendu:
is this branch align with `main` branch? Because I found some changes which is already present in main.
Please align this branch with `main`.
Also I feel this PR is for publish result. is run CI from `snappi` development branch also include in this PR.
It will better to put some descriptions the feature we are addressing.

PR #580, opened 2025-01-13
Pull otg model

indraniBh:
@arkajyoti-cloud @alakendu @dipendughosh


version-master.zip
indraniBh, line 1: Please remover the binary file

version.py
indraniBh, line 1: What is the purpose of this file?

indraniBh (commented):
@desaimg1 Please take care of the given comments.

indraniBh:
@desaimg1 I can see there is conflict in do.py with main. Please sync it with latest main.

Vibaswan:
This looks fine 
but i have certain questions 
So are we thinking snappi Ci totally working independently without the snappi repo ?
I am asking this because okay only models will then suffice , but it creates dependency on openapiart which does not make much sense for me for snappi-ixn repo
instead of generating snappi in pipeline itself, cant we just take mention snappi branch instead of models branch (because eventually one needs to create snappi branch right to check the models) , so in that case we can directly get the snappi python file and source it instead of building the whole thing in the ci.

@alakendu  I think otg-hw just gets the snappi file right 
@desaimg1 and @indraniBh is there a particular reason why we went this way ?

desaimg1:
> This looks fine but i have certain questions So are we thinking snappi Ci totally working independently without the snappi repo ? I am asking this because okay only models will then suffice , but it creates dependency on openapiart which does not make much sense for me for snappi-ixn repo instead of generating snappi in pipeline itself, cant we just take mention snappi branch instead of models branch (because eventually one needs to create snappi branch right to check the models) , so in that case we can directly get the snappi python file and source it instead of building the whole thing in the ci.
> 
> @alakendu I think otg-hw just gets the snappi file right @desaimg1 and @indraniBh is there a particular reason why we went this way ?

Hi Vibaswan,

    In otg-hw i see it just importing snappi from particular branches. If any changes made in models do we need to create an snappi branch manually and then import in our snappi-ixnetwork ci?
 
Thanks,
Desai

Vibaswan:
> > This looks fine but i have certain questions So are we thinking snappi Ci totally working independently without the snappi repo ? I am asking this because okay only models will then suffice , but it creates dependency on openapiart which does not make much sense for me for snappi-ixn repo instead of generating snappi in pipeline itself, cant we just take mention snappi branch instead of models branch (because eventually one needs to create snappi branch right to check the models) , so in that case we can directly get the snappi python file and source it instead of building the whole thing in the ci.
> > @alakendu I think otg-hw just gets the snappi file right @desaimg1 and @indraniBh is there a particular reason why we went this way ?
> 
> @Vibaswan In otg-hw i see it just importing snappi from particular branches. If any changes made in models do we need to create an snappi branch manually and then import in our snappi-ixnetwork ci?

Yes that's what the usual practice is, whenever a individual does model changes in order for using it, they always create a branch in snappi (manually) for everyone to use.
I understand your approach to totally make it independent of the snappi repo , but snappi ci actually runs a lot of sanity tests when we create a branch which is missing in this workflow, hence if a bug arises in snappi itself , you guys might confuse it being something related to snappi-ixn and spend more time finding it.

PR #581, opened 2025-01-14
CP-DP convergence support in snappi-ixnetwork

indraniBh:
This PR contains:
**Implementation:**
1. Support for CP-DP convergence metrics as part of snappi-ixnetwork using core OTG model.
2. Support of accepting username/password for convergence [as it is part of snappi-ixnetwork now]
3. Port_metrics and flow_metrics support for convergence TCs

**Tests:**
1. Update existing convergence tests exercising different supported events for multiple route [route_advertise, route_withdraw, link_up and link_down]
2. ping support
3. DP only TC
4. Multi Rx port multi Route TC [Skipped for CI]

FYI @arkajyoti-cloud @alakendu


snappi_ixnetwork/snappi_api.py
```diff
@@ -669,6 +637,225 @@ def update_flows(self, payload):
...
+                raise Exception("Somehow flow %s is missing" % flow_name)
+
+            drill_down_options = traffic_stat.DrillDownOptions()
+            drilldown_index = drill_down_options.index(drill_down_option)
+            drill_down = traffic_stat.Drilldown(
+                drilldown_index,
+                drill_down_option,
+                traffic_stat.TargetRowFilters()[traffic_index[flow_name]],
+            )
+            drill_down_result = self._get_max_convergence(drill_down.Rows)
```
alakendu, line 697: expecting drill down happen before that. It will better to check Row has some length.

alakendu (approved):
made some small comments. Please take a look

snappi_ixnetwork/snappi_api.py
```diff
@@ -669,6 +637,225 @@ def update_flows(self, payload):
...
+                        self._set_result_value(
+                            convergence,
+                            "service_interruption_time_us",
+                            interruption_time,
+                            float,
+                        )
+                convergence["events"] = events
+
+            # for DP only metric
+            if ixn_cpdpconvergence.EnableDataPlaneEventsRateMonitor and not ixn_cpdpconvergence.EnableControlPlaneEvents: # noqa
```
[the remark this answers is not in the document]
indraniBh: Yes. Two use cases, cp-dp and dp only. In CP only there is no stats CP/CP convergence time. 
There are only two DP/DP convergence and CP/DP convergence.

snappi_ixnetwork/snappi_api.py
indraniBh: Done.

PR #582, opened 2025-01-17
Add support for missing fields under FilterPallete to add capture fil…

dipendughosh:
…ters.

### DIFF
--- a/snappi_ixnetwork/capture.py
+++ b/snappi_ixnetwork/capture.py
@@ -5,6 +5,12 @@ from snappi_ixnetwork.timer import Timer
 from snappi_ixnetwork.logger import get_ixnet_logger
 
 
+class CommonField:
+    def __init__(self, field, offset):
+        self.field = field
+        self.offset = offset
+
+
 class Capture(object):
     """Transforms OpenAPI objects into IxNetwork objects
 
@@ -73,7 +79,6 @@ class Capture(object):
                         "xpath": vports[port_name]["xpath"]
                         + "/capture/trigger"
                     }
-                    # reset["captureTriggerEnable"] = False
                     self._import(reset)
                 capture = {
                     "xpath": vports[port_name]["xpath"] + "/capture",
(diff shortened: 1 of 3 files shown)


snappi_ixnetwork/capture.py
```diff
@@ -155,57 +179,395 @@ def _config_missing_pallete(self, cap_filter, pallette, trigger, filter):
...
-            if (
-                cap_filter.src.negate is not None
-                and cap_filter.src.negate is True
-            ):
+        return expressionString
+
+    def _config_ethernet_pallette(
+        self, cap_filter, pallette, trigger, filter, expressionString
+    ):
+        if not cap_filter.src.value == "00":
```
alakendu, line 231: We are not setting any value if user is setting `00`. This is not a valid case for `src`, but that will valid for other field.
I feel current python snappi do not have ready solution.
What about if we check with `cap_filter.src.value is not None` , is it `True` all the times.
We can try this one as well: https://github.com/open-traffic-generator/snappi-ixnetwork/blob/main/snappi_ixnetwork/validation.py#L34

snappi_ixnetwork/capture.py
```diff
@@ -155,57 +179,395 @@ def _config_missing_pallete(self, cap_filter, pallette, trigger, filter):
...
+            expressionString = self._config_source_address(
+                cap_filter.src, pallette, trigger, filter, expressionString
+            )
+        if not cap_filter.dst.value == "00":
+            expressionString = self._config_destination_address(
+                cap_filter.dst, pallette, trigger, filter, expressionString
+            )
+        if not cap_filter.ether_type.value == "00":
```
alakendu, line 239: This is not a must change. 
calling multiple `_config_common_filter_pallete` is little annoying. is it not possible to maintain a class has all those property in a dics. Then loop through and call  `_config_common_filter_pallete` once.

snappi_ixnetwork/capture.py
```diff
@@ -155,57 +179,395 @@ def _config_missing_pallete(self, cap_filter, pallette, trigger, filter):
...
                 filter["captureFilterDA"] = "addr1"
-            trigger["triggerFilterDA"] = filter["captureFilterDA"]
+            trigger["captureTriggerDA"] = filter["captureFilterDA"]
+            if len(expressionString) == 0:
+                expressionString = negate + "DA1"
+            else:
+                expressionString = expressionString + " and " + negate + "DA1"
+        return expressionString
+
+    def _config_common_filter_pallete(
```
alakendu, line 524: This is not a must change.
calling multiple _config_common_filter_pallete is little annoying. is it not possible to maintain a class has all those property in a dics. Then loop through and call _config_common_filter_pallete once.

snappi_ixnetwork/capture.py
```diff
@@ -155,57 +179,395 @@ def _config_missing_pallete(self, cap_filter, pallette, trigger, filter):
...
+            expressionString = self._config_common_filter_pallete(
+                cap_filter.ether_type,
+                pallette,
+                trigger,
+                filter,
+                12,
+                expressionString,
+            )
+        if not cap_filter.pfc_queue.value == "00":
+            expressionString = self._config_common_filter_pallete(
```
alakendu, line 249: Please validate `pfc_queue` is working. I did not find it from IxNetwork Manager GUI

snappi_ixnetwork/capture.py
```diff
@@ -155,57 +179,395 @@ def _config_missing_pallete(self, cap_filter, pallette, trigger, filter):
...
+    def _config_vlan_pallette(
+        self, cap_filter, pallette, trigger, filter, expressionString
+    ):
+        if not cap_filter.priority.value == "00":
+            expressionString = self._config_common_filter_pallete(
+                cap_filter.priority,
+                pallette,
+                trigger,
+                filter,
+                0,
```
alakendu, line 268: I have some confusion about `offset`. Could you please check only VLAN filer, then Ethernet and VLAN filter.

snappi_ixnetwork/capture.py
dipendughosh: taken care off in the next commit.

snappi_ixnetwork/capture.py
dipendughosh: taken care off in the next commit.

snappi_ixnetwork/capture.py
dipendughosh: taken care off in the next commit.

PR #588, opened 2025-02-06
fix: set /topology/deviceGroup/ethernet/enableVlans as multivalue

mtache:
`/topology/deviceGroup/ethernet/enableVlans` was not correctly set by snappi, the device interface never get the enableVlans set to `True`.
I fixed the code to use multivalue instead which is expected by IxNetwork and I wrote a test to cover this issue.


arkajyoti-cloud:
Thanks for pointing it. we will review and take it in.

indraniBh (approved):
LGTM

PR #591, opened 2025-02-21
MACsec/ MKA config and stats support along with tests in snappi-ixnetwork

sasubrata:
MACsec/ MKA config and stats support along with tests in snappi-ixnetwork


tests/mka/test_mka_stats.py
```diff
@@ -0,0 +1,126 @@
+import pytest
```
alakendu, line 1: Somehow snappi_ixnetwork group mail I have missed as that is way down within Group. Anyway, best way to check this huge feature to test all field. I will check these changes. 
At the same time in slack I have mentioned - snappi_ixnetwork properly compacted MacSec in topology level.
Create multiple similar OTG MacSec config within a single port. All IxNetwork MacSec  DG should compacted to single DG with proper multiplier and MV.

sasubrata:
Hi Alakendu,

I understand that only review comment from your side is to test all fields and also test with multiple devices in OTG.

AFAIK, there is no “device group” in OTG – but there can be multiple devices.

Regards,
Subrata

snappi_ixnetwork/device/macsec.py
```diff
@@ -0,0 +1,333 @@
...
+            ixn_staticmacsec["incrementingPn"] = False
+            engine_tx_pn_fixed = engine_tx_pn.fixed
+            ixn_staticmacsec["fixedPn"] = engine_tx_pn_fixed.pn
+            ixn_staticmacsec["mvFixedXpn"] = self.multivalue(engine_tx_pn_fixed.xpn)
+        elif engine_tx_pn.choice == "incrementing_pn":
+            ixn_staticmacsec["incrementingPn"] = True
+            engine_tx_pn_incr = engine_tx_pn.incrementing
+            ixn_staticmacsec["packetCountPn"] = engine_tx_pn_incr.count
+            ixn_staticmacsec["firstPn"] = engine_tx_pn_incr.first_pn
+            ixn_staticmacsec["mvFirstXpn"] = self.multivalue(engine_tx_pn_incr.first_xpn)
```
alakendu, line 333: not found any code to change card mode to masec otherwise user will raise issue if we want their card should be converted to macsec mode.

tests/mka/test_mka_stats.py
```diff
@@ -0,0 +1,126 @@
...
+import time
+
+
+# @pytest.mark.skip(reason="Revisit CI/CD fail")
+def test_mka_stats(api, b2b_raw_config, utils):
+    """
+    Test for the mka configuration
```
alakendu, line 8: It will better to add test for field validation. Use any of the config then made `set_config` finally need to validate those field are setting properly within IxNetwork. Please referred this test https://github.com/open-traffic-generator/snappi-ixnetwork/blob/main/tests/bgp/test_bgp_attributes.py#L128

PR #592, opened 2025-02-27
Chassis Chain support for IxNetwork only

dipendughosh:


snappi_ixnetwork/ixnetworkconfig.py
```diff
@@ -0,0 +1,190 @@
...
+            self._chassis_chains = ChassisChainIter()
+    
+        return self._chassis_chains
+
+    def config(self):
+        self.logger.info("Configuring IxNConfig")
+
+class ChassisChain(object):
+    _TYPES = {
+        "primary": {"type": str},
```
alakendu, line 30: It will good if we add docstring for all user configurable property similar as `snappi`. It will maintain same Specification as `sanppi` also help user to write script.

tests/chassischain/test_chassis_chain_support_2.py
```diff
@@ -0,0 +1,126 @@
+def test_chassis_chain_support_2(api, b2b_raw_config, utils):
+
+    # Chassischain configuration
+    ixnconfig = api.ixnetworkconfig
```
alakendu, line 4: we need to add validation that those are setting properly through IxNetwork. This is RestPy IxNetwork handler 'api._ixnetwork`. So, you will get chassis using `api._ixnetwork.AvailableHardware.Chassis`. Please take a look in this test for reference `tests\vxlan\test_manual_gateway_mac.py` 


These tests probably not run in regular pipeline.

alakendu (approved):

PR #599, opened 2025-03-26
Support for MACsec flow header in order to shaping MACsec traffic payload

sasubrata:
Add support for MACsec header so that payload IP DSCP properties can be set using flow.packet.ethernet().macsec().ipv4() sequence.

### DIFF
--- a/snappi_ixnetwork/trafficitem.py
+++ b/snappi_ixnetwork/trafficitem.py
@@ -66,6 +66,8 @@ class TrafficItem(CustomField):
         "custom": "custom",
         "vxlan": "vxlan",
         "ethernetARP": "arp",
+        "macsec": "macsec",
+        "payloadProtocolType": "payloadProtocolType",
     }
 
     _HEADER_TO_TYPE = {
@@ -83,6 +85,8 @@ class TrafficItem(CustomField):
         "custom": "custom",
         "vxlan": "vxlan",
         "arp": "ethernetARP",
+        "macsec": "macsec",
+        "payloadProtocolType": "payloadProtocolType",
     }
 
     _ETHERNETPAUSEUHD = {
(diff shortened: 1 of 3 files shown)


snappi_ixnetwork/trafficitem.py
```diff
@@ -912,41 +919,44 @@ def _configure_packet(self, ixn_stack, snappi_packet):
...
-                self._append_header(header, xpath, stacks)
+                self._append_header(xpath, stacks, header)
         return stacks
 
     def _append_header(
         self,
-        snappi_header,
         xpath,
         stacks,
+        snappi_header=None,
```
alakendu, line 937: I feel we are not made any operation if  `snappi_header=None`  then we can continue from the calling loop without called `_append_header`.

tests/macsec/test_macsec_mka_traffic.py
```diff
@@ -135,6 +135,20 @@ def test_encrypt_with_mka(api, b2b_raw_config, utils):
     # Flow
     f1 = config.flows.flow(name="f1")[-1]
 
+    # IP
```
alakendu, line 138: We need to validate the RAW scenario when packets has `macsec` e,g, `f1.packet.ethernet().macsec().ipv4()` . 
```
    flow.tx_rx.port.tx_name = tx.name
    flow.tx_rx.port.rx_name = rx.name
```
Then it will call `_append_header` to configure `macsec`. We need to check we have missed something in that path.

alakendu (approved):

snappi_ixnetwork/trafficitem.py
sasubrata: "PayloadProtocolType" header is appended without snappi_header argument i.e. snappi_header=None because we do not expose this header in snappi (OTG model)

indraniBh:
Should we not update snappi version to 1.27.0?

PR #603, opened 2025-04-28
RoCEv2 Support in Snappi

satyamsinghKey:
Adding RoCEv2 Support.

Test Cases added:
test_rocev2_stats.py
test_rocev2_traffic.py
test_rocev2_stats_multiplesQPs.py

NOTE: These test cases are marked for skip, as these require RoCEv2 specific ports and will fail without them.

### DIFF
--- a/do.py
+++ b/do.py
@@ -11,7 +11,7 @@ from email.mime.text import MIMEText
 
 global ixnexception
 
-SNAPPI_BRANCH=None
+SNAPPI_BRANCH="dev2_rocev2"
 
 def get_snappi_dev_branch():
     if SNAPPI_BRANCH is not None and SNAPPI_BRANCH != "":
--- a/do.py
+++ b/do.py
@@ -11,7 +11,7 @@ from email.mime.text import MIMEText
 
 global ixnexception
 
-SNAPPI_BRANCH="dev2_rocev2"
+SNAPPI_BRANCH="dev2_roce2"
 
 def get_snappi_dev_branch():
     if SNAPPI_BRANCH is not None and SNAPPI_BRANCH != "":
--- a/do.py
+++ b/do.py
@@ -11,7 +11,7 @@ from email.mime.text import MIMEText
 
 global ixnexception
 
-SNAPPI_BRANCH="dev2_roce2"
+SNAPPI_BRANCH=None
 
 def get_snappi_dev_branch():
     if SNAPPI_BRANCH is not None and SNAPPI_BRANCH != "":
--- a/snappi_ixnetwork/device/rocev2.py
+++ b/snappi_ixnetwork/device/rocev2.py
@@ -8,9 +8,6 @@ class RoCEv2(Base):
     }
 
     _MANDATE_CONFIG = {
-        "source_qp_number": "customQP",
-        "dscp": "dscp",
-        "udp_source_port": "udpSourcePort",
         "initial_psn": "initialPSN",
         "virtual_address": "remoteVA",
         "remote_key": "remoteKey"
@@ -42,10 +39,6 @@ class RoCEv2(Base):
         }  
     }
 
-    _GLOBALS = {
-        "message_size": "CnpPriorityValue"
-    }
-
     ecn_mapping = {
         "non_ect": 0, "ect_1": 1, "ect_0": 2, "ce": 3
     }
@@ -57,13 +50,21 @@ class RoCEv2(Base):
         self._ngpf = ngpf
         self.logger = get_ixnet_logger(__name__)
 
-    def config(self, device, stateful_flow, options):
+    def config(self, device):
         self.logger.debug("Configuring RoCEv2")
         rocev2 = device.get("rocev2")
         if rocev2 is None:
             return
+        stateful_flow = None
+        options = None
+        if hasattr(self._ngpf.api.snappi_config, "stateful_flows"):
+            stateful_flow = self._ngpf.api.snappi_config.stateful_flows
+        if hasattr(self._ngpf.api.snappi_config, "options"):
+            options = self._ngpf.api.snappi_config.options
         self._config_ipv4_interfaces(rocev2, stateful_flow, options)
         self._config_ipv6_interfaces(rocev2, stateful_flow, options)
+
+        
 
     def _get_interface_info(self):
         ip_types = ["ipv4", "ipv6"]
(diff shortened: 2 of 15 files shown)


do.py
```diff
@@ -11,7 +11,7 @@
 
 global ixnexception
 
-SNAPPI_BRANCH=None
+SNAPPI_BRANCH="dev2_roce2"
```
indraniBh, line 14: Revert it.

tests/settings.json
indraniBh, line 1: Please change settings.json to default mode.

tests/rocev2/test_rocev2_traffic.py
indraniBh, line 1: For all test files to be skipped, please site the proper reason for exclusion in CI.
IMO, " Skipping this test in CI due to setup unavailability" .

indraniBh (changes requested):
I have given some comments mostly on infrastructure side. Please take care of it.

tests/settings.json
satyamsinghKey: done

tests/rocev2/test_rocev2_traffic.py
satyamsinghKey: done

indraniBh (approved):

indraniBh (approved):
LGTM from infra side.

snappi_ixnetwork/device/rocev2.py
```diff
@@ -0,0 +1,337 @@
...
+        trafficPortConfigs = rocev2_traffic.RoceV2PortConfig.find()
+        trafficflows = rocev2_traffic.RoceV2Stream.find()
+        rocev2s = []
+        if (stateful_flow is not None):
+            rocev2s = stateful_flow.get("rocev2")
+        for trafficportconfig in trafficPortConfigs:
+            for rocev2 in rocev2s:
+                for tx_port in rocev2.tx_ports:
+                    if (trafficportconfig.TxPort == tx_port.port_name):
+                        trafficportconfig.TargetLineRateInPercent = tx_port.transmit_type.target_line_rate.value
```
alakendu, line 301: are we not supporting  `Rocev2.TargetLineRate.Flow`. Because not find `tx_endpoint` and `rx_endpoint` within the implementation though test have those fields.

snappi_ixnetwork/device/rocev2.py
```diff
@@ -0,0 +1,337 @@
...
+                        if ecn_numeric is not None:
+                            ixnRocev2GlobalPortSettings.NakEcnVal.Single(ecn_numeric)
+                    #Retransmission        
+                    # ixnRocev2GlobalPortSettings.EnableACKTimeout.Single(protocol.connection_type.reliable_connection.get("enable_retransmission_timeout"))
+                    # ixnRocev2GlobalPortSettings.AckTimeout.Single(protocol.connection_type.reliable_connection.get("retransmission_timeout_value"))
+                    # ixnRocev2GlobalPortSettings.RetransRetryCount.Single(protocol.connection_type.reliable_connection.get("retransmission_retry_count"))
+
+    def _configureTrafficParameters(self, rocev2_traffic, stateful_flow=None, options=None):
+        trafficPortConfigs = rocev2_traffic.RoceV2PortConfig.find()
+        trafficflows = rocev2_traffic.RoceV2Stream.find()
```
alakendu, line 293: We have a Four port test  (Two Tx and Two Rx ports) but in the stateful_flows user is configured only single `tx_ports.port_name`. User expectation flow should create within two roce port. But, it looks like we are creating flows for all the ports. Please double check if I have missing anything.

alakendu (changes requested):

snappi_ixnetwork/device/rocev2.py
```diff
@@ -0,0 +1,337 @@
...
+        for trafficportconfig in trafficPortConfigs:
+            for rocev2 in rocev2s:
+                for tx_port in rocev2.tx_ports:
+                    if (trafficportconfig.TxPort == tx_port.port_name):
+                        trafficportconfig.TargetLineRateInPercent = tx_port.transmit_type.target_line_rate.value
+
+        for trafficportconfig in trafficPortConfigs:
+            dcqcn_params = trafficportconfig.RoceV2DcqcnParams
+            for dcqcn_param in dcqcn_params:
+                if (options is not None):
```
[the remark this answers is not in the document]
satyamsinghKey: done

snappi_ixnetwork/trafficitem.py
```diff
@@ -1330,6 +1385,11 @@ def transmit(self, request):
                     )
             self._api.capture._start_capture()
         self._api._traffic_item.find(Name=regex)
+        for device in self._api._config.devices:
```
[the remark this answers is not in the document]
satyamsinghKey: I have added a check to address the issue, now traffic will only stop when stateful_flows are configured in script. Also, if the user want to disable a flow, they can simply disble the flow in the script.

tests/rocev2/test_rocev2_stats_multipleQPs.py
```diff
@@ -0,0 +1,162 @@
...
+    )
+
+
+    print ("Stopping Protocols")
+    control_state = api.control_state()
+    control_state.protocol.all.state = control_state.protocol.all.STOP
+    api.set_control_state(control_state)
+
+def results_ok(api):
+    req = api.metrics_request()
```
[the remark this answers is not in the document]
satyamsinghKey: Yes, I have done that in my local dev setup, printing all the stats and verifying.

snappi_ixnetwork/device/rocev2.py
satyamsinghKey: tx_endpoint is used to map correct qp number to correct row in ixnetwork. Since its not like other traffic items, here the flows are automatically generated and we have no control of endpoints. IN our case, tx_endpoint is being used internally to identify to which QP does the message fields will be populated.

snappi_ixnetwork/device/rocev2.py
satyamsinghKey: This is a limitation from IxNetwork end. Where when the user does "Add L2-3 RoCEv2 Flow Groups" the flows will be generated automatically for all the QPs configured on all the ports. Hence the same behavior is here as well. However, the model should be generic enough to accommodate all the use cases.

alakendu (approved):

PR #606, opened 2025-05-03
CI Bug Fix

indraniBh:
1. Bug fix: Fix test coverage logic
2. Added Pass rate
3. Island failure if UT or sanity pass rate fall under threshold
4. Mail report on each submission only in main.


arkajyoti-cloud:
Approved

PR #611, opened 2025-05-18
Append and delete config (flows)

indraniBh:


snappi_ixnetwork/trafficitem.py
```diff
@@ -2157,4 +2157,277 @@ def rocev2_flow_results(self, request):
...
+        start_states = [
+            "txStopWatchExpected",
+            "locked",
+            "started",
+            "startedWaitingForStats",
+            "startedWaitingForStreams",
+            "stoppedWaitingForStats",
+        ]
+        state = self._api._ixnetwork.Traffic.State
+        if state in start_states:
```
alakendu, line 2177: is it require tom stop all the flows.

indraniBh: Yes.

snappi_ixnetwork/trafficitem.py
```diff
@@ -2157,4 +2157,277 @@ def rocev2_flow_results(self, request):
...
+            self._api._ixnetwork.Traffic.StopStatelessTrafficBlocking()
+
```
alakendu, line 2179: According to our Athena-G discussion, We were decide to check that the traffic actually stopped.

indraniBh: For implicit call we are checking whether flows are in started state or not, if it is we stopped all the traffic before append/delete flows otherwise it will throw error from ixnetwork.

snappi_ixnetwork/trafficitem.py
```diff
@@ -2157,4 +2157,277 @@ def rocev2_flow_results(self, request):
...
+            initial_flownames.append(i_flow.name)
+        errors = []
+        for flow in append_flows_config:
+            if flow.name in initial_flownames:
+                errors.append(
+                    "flow {} is already present in the configuration".format(
+                        flow.name
+                    )
+                )
+            self._api._config.flows._items.append(flow)
```
alakendu, line 2430: probably we need to put it into else block otherwise that flow actually added though we have error on it.

tests/traffic-properties/test_append_flow.py
```diff
@@ -0,0 +1,108 @@
...
+
+    config = api.get_config()
+    print(config)
+
+    cd = api.config_delete()
+    cd.config_delete_list.add().flows = ["flw2"]
+    print("Deletion request for the flows", cd)
+    api.delete_config(cd)
+
+    config = api.get_config()
```
alakendu, line 104: please add condition in `get_config` after append or delete

tests/traffic-properties/test_delete_flows.py
```diff
@@ -0,0 +1,99 @@
...
+
+    # Start traffic
+    control_state = api.control_state()
+    control_state.choice = control_state.TRAFFIC
+    control_state.traffic.choice = control_state.traffic.FLOW_TRANSMIT
+    control_state.traffic.flow_transmit.state = control_state.traffic.flow_transmit.START  # noqa
+    api.set_control_state(control_state)
+    time.sleep(5)
+
+    utils.get_all_stats(api)
```
alakendu, line 85: try to get_metrics using the deleted flow. snappi_ixnetwork should not return any metric which already deleted.

indraniBh: Done.

PR #612, opened 2025-05-20
Support for MKA pre-shared key(PSK) chain start time in UTC time format

sasubrata:
Feature description
===============
User can configure MKA pre-shared key(PSK) chain start time in UTC time format in MACsec device when key generation protocol is MKA. If the value as per models path device->macsec->ethernet_interfaces->secure_entity->key_generation_protocol->mka->basic->psk_chain_start_time is configured for one device, the same value is used globally across ports for all such devices.

Use-cases
===============
User can match PSK chain start time with that of DUT

Added TCs with brief description
=====================
tests/mka/test_mka_stats.py:
...
    kay1.basic.psk_chain_start_time.choice = kay2.basic.psk_chain_start_time.choice = "utc"
    kay1_psk_chain_start_time = kay1.basic.psk_chain_start_time.utc
    kay2_psk_chain_start_time = kay2.basic.psk_chain_start_time.utc

    kay1_psk_chain_start_time.day = kay2_psk_chain_start_time.day = 6
    kay1_psk_chain_start_time.month = kay2_psk_chain_start_time.month = 5
    kay1_psk_chain_start_time.year = kay2_psk_chain_start_time.year = 2025
    kay1_psk_chain_start_time.hour = kay2_psk_chain_start_time.hour = 7
    kay1_psk_chain_start_time.minute = kay2_psk_chain_start_time.minute = 10
    kay1_psk_chain_start_time.second = kay2_psk_chain_start_time.second = 56
...


snappi_ixnetwork/snappi_api.py
```diff
@@ -368,6 +372,16 @@ def config_ixnetwork(self, config):
             # Start all protocols is workaround for pCPU crash reported by
             # Microsoft, need to revert once fix is available for pCPU crash
             if self._protocols_exists():
+                for device in self._config.devices:
+                    #After the config is pushed in IxNetwork, only then we can do ClearOverlays() in MKA Globals.
+                    #That is why it cannot be done from within MKA module itself.
+
+                    #Check if MACsec/MKA exists in topology
+                    macsec = device.get("macsec")
```
alakendu, line 380: is it require to loop multiple times when we found one `macsec` . Because we are clear the overlays. So, it will clear for all in the first call. We can break after execution of the first execution.

snappi_ixnetwork/device/mka.py
```diff
@@ -158,6 +158,52 @@ def _config_basic(self, kay, ixn_mka):
...
+                        return True
+        return False
+                    
+    def _clear_overlays_in_globals(self, macsec):
+        if (self._is_test_start_time_user_defined(macsec)):
+            ixn_mka_globals_port_settings = self._ngpf._ixnetwork.Globals.Topology.find().Mka.find()
+            ixn_mka_globals_port_settings.TestStartTime.ClearOverlays()
+
+    def _config_test_start_time(self, basic, ixn_mka):
+        self.logger.debug("Configuring test start time")
```
alakendu, line 185: I feel this global `TestStartTime` is per port. But we are setting it for per OTG device. What about if user configure two different time within two MKA present in same port.

sasubrata: In one connectivity association (CA), MKA properties of all devices should be the same - so test start time (which is actually key chain start time) is per CA. Now a test tool can emulate multiple CAs per physical port - Ixia can do so. Although Ixia allows setting test start time per physical port, some other test tool vendor may allow setting the time per CA e.g. per device or group of devices in emulated CA.

Hence the test start time is kept per device in OTG model. Test tool vendor e.g. snappi-ixnetwork should validate whether the values are the same for all devices per CA. In snappi-ixnetwork, currently this validation is not done - it can be added later.

snappi_ixnetwork/snappi_api.py
sasubrata: There can be two MACsec devices. Each device can be assigned different physical port. Now the test start time property is per physical port. Hence we need to 1) do it per device when each device is assigned different physical port and additionally we can 2) keep track that we we not doing clear overlays multiple time per physical port when multiple devices are mapped to same physical port.

While the looping is done to take care of scenario 1) above, optimisation can be done to take care of scenario 2) later.

PR #626, opened 2025-09-19
New build strategy in python

indraniBh:


pyproject.toml
Vibaswan, line 1: tool.black can still be there i do not think we need to remove this

do.py
```diff
@@ -447,23 +452,6 @@ def get_workflow_id():
     return workflow_id
 
 
-def check_release_flag(release_flag=None, release_version=None):
```
Vibaswan, line 450: is this not needed at all ?

Vibaswan (approved):
overall LGTM!

PR #628, opened 2025-09-23
Multiple ipv6 route range  support for convergence 

indraniBh:
**Bug Description:** Route Withdraw with heterogenous route ranges attached to the same bgp stack is not working.
**Resolution:** router name of BGPv6 route was not set properly
**Added TC:** Convergence TC with v6 multiple-route ranges

FYI @arkajyoti-cloud


alakendu (approved):

indraniBh:
TODO: Modify test with bgpv6 attributes validation (route name)

PR #647, opened 2026-04-28
mpls support

indraniBh:


snappi_ixnetwork/trafficitem.py
```diff
@@ -204,6 +206,14 @@ class TrafficItem(CustomField):
 
     _MACSEC = {}
 
+    _MPLS = {
+        "label": "mpls.label.value",
```
alakendu, line 210: I believe the most critical section for MPLS is label.auto, as it automatically populates labels for device traffic. However, it’s possible that snappi_ixnetwork is using device endpoints instead—please double-check this.
Also we need to validate those, check one such test case [with RSVP](https://nts-gitlab.is.keysight.com/athena/otg-ixhw-c/-/blob/main/pytest/otg_tests/system_tests/test_rsvp_bidirectional_traffic.py?ref_type=heads) ,  
A similar scenario should apply to SRv6. Additionally, ensure that the label is correctly reflected in the packet and also verified in the tracking fields.

indraniBh: RSVP is not implemented here yet in snappi-ixnetwork. 
While we implement SRv6 it will be verified, will do the changes while srv6 in place.

tests/mpls/test_mpls_header.py
```diff
@@ -0,0 +1,115 @@
+def test_mpls_header(api, b2b_raw_config_vports, utils, tx_vport, rx_vport):
+    """
+    Configure three raw MPLS flows with ethernet + mpls + ipv4 stack:
+    - f1: fixed label, traffic_class, bottom_of_stack, time_to_live
+    - f2: list pattern for label and time_to_live
+    - f3: increment pattern for label, decrement pattern for time_to_live
```
alakendu, line 6: Also validate multiple MPLS level

indraniBh: ok

indraniBh:
These is two use case:
1. Only MPLS Traffic where we put predefined MPLS level stack and DUT will work accordingly,
2. SRv6 + MPLS Traffic where SRV6 control plan first communicate with DUT to established session then MPLS Traffic flow

Revisit this part when srv6 in place.

tests/mpls/test_mpls_header.py
indraniBh: Done

PR #658, opened 2026-06-29
Bidirectional traffic and Frame ordering support in flow

indraniBh:
Added bidirectional support in flow, if this is enabled  in traffic item, creates traffic sub-flows on both the forward (tx_names->rx_names) and reverse (rx_names->tx_names) directions.

Added Frame ordering Feature.
Options->per_port_options-> frame_ordering (choices: no_ordering (default behaviour) and RFC2889)

Test Case:
1. Enable bidirectional flow and validate it in traffic item whether it is enabled or not.
2. Validate default value as false in Traffic item if it's not given.
3. Added Full mesh test using bi-directional traffic and RFC2889

### DIFF
--- a/snappi_ixnetwork/trafficitem.py
+++ b/snappi_ixnetwork/trafficitem.py
@@ -722,6 +722,7 @@ class TrafficItem(CustomField):
                     "xpath": tr_xpath,
                     "name": "%s" % flow.name,
                     "srcDestMesh": self._get_mesh_type(flow),
+                    "biDirectional": self._get_bidirectional(flow),
                 }
             )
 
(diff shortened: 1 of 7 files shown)


snappi_ixnetwork/trafficitem.py
```diff
@@ -1337,6 +1338,17 @@ def _get_mesh_type(self, flow):
...
 
+    def _get_bidirectional(self, flow):
+        """Return whether the flow's device endpoints request bidirectional
+        traffic. When enabled, IxNetwork creates traffic sub-flows on both the
+        forward (tx_names -> rx_names) and reverse (rx_names -> tx_names)
+        directions. Only device endpoints support this; port endpoints are
+        always unidirectional.
+        """
+        if flow.tx_rx.choice == "device":
+            return bool(flow.tx_rx.device.bidirectional)
```
alakendu: Add a None validation to ensure that the configuration includes bidirectional settings.

alakendu (approved):

snappi_ixnetwork/trafficitem.py
indraniBh: Done.

dipendughosh-k (approved):

PR #666, opened 2026-08-11
Support get_states with bgp_prefixes (BGP learned-route retrieval)

indraniBh:
**Summary**
Implements the **bgp_prefixes** choice of the OTG **StatesRequest/StatesResponse** API on the **IxNetwork backend.** Previously get_states only accepted ipv4_neighbors and ipv6_neighbors; callers asking for learned BGP routes got a TypeError. This PR adds end-to-end support for querying learned **IPv4/IPv6 unicast prefixes** from configured BGP peers and translating them into OTG-shaped BgpPrefixesState objects.

**What changed**
[snappi_ixnetwork/device/ngpf.py]

- Routes request.choice == "bgp_prefixes" to a new get_bgp_prefix_states().
- Resolves the requested peer names to live IxNetwork peer objects, fetches learned prefixes per peer session, and returns {"choice": "bgp_prefixes", "bgp_prefixes": [...]} with results keyed under ipv4_unicast_prefixes / ipv6_unicast_prefixes by family.
- Updated the unsupported-choice error message accordingly.

[snappi_ixnetwork/device/bgp.py]
- **Config**: learned_information_filter on a BGP peer is now applied as IxNetwork multivalues (filterIpV4Unicast / filterIpV6Unicast) for both BGPv4 and BGPv6 peers. Without at least one filter enabled, IxNetwork stores nothing in learnedInfo, so this is required for the retrieval path to return data.
- **Peer resolution** (get_bgp_peer_objects): walks Topology → DeviceGroup → Ethernet → Ipv4/Ipv6 → BgpIpv4Peer/BgpIpv6Peer and maps snappi peer names to (name, peer_obj, 1-based session index, family). Handles both non-compacted single-session peers and compacted multi-session scalable peers (session index derived from the name's position in the ixn_objects names list). An empty bgp_peer_names means all peers; unknown names raise.
- **Table retrieval** (_get_learned_table): triggers GetAllLearnedInfo() and reads LearnedInfo.find().Table.find(). Note the deliberate choice of GetAllLearnedInfo over GetIPv4/6LearnedInfo — the latter only updates the deprecated inline Columns/Values fields that IxNetwork 10.x no longer writes.
- **Row → OTG translation** (_row_to_ipv4_prefix / _row_to_ipv6_prefix): column-name lookups use ordered alias tuples so the code tolerates IxNetwork version-to-version naming drift. Handles address columns that carry either a bare address plus a separate prefix-length column or a full CIDR string, distinguishes IPv4 vs IPv6 next hops (including the legacy single Next Hop column), and populates origin, local_preference, multi_exit_discriminator, and path_id only when the source cell holds a real value (N/A/empty/0 are skipped).
- **AS-path parser** (_parse_as_path): converts IxNetwork AS-path strings into OTG segment lists, including {...} AS_SET, (...) AS_CONFED_SEQ, and [...] AS_CONFED_SET groupings interleaved with plain AS_SEQ runs.
- **Community parser** (_parse_communities): handles as:custom pairs (tolerating whitespace around the colon) and well-known values (no-export, no-advertise, no-export-subconfed, llgr_stale, no_llgr); unrecognised tokens are logged and skipped rather than raising.
- **Filtering** (_apply_v4_filters / _apply_v6_filters): applies ipv4_unicast_filters / ipv6_unicast_filters from the request. Filters are OR-ed with each other; fields within a filter (addresses, prefix_length, origin, path_id) are AND-ed, and unset fields act as wildcards. No filters means no restriction.

**Tests**
[tests/bgp/test_bgp_prefix_states.py]

- **test_bgp_prefix_states_ipv4_b2b** — verifies learned IPv4 prefixes, spot-checks address/prefix-length/next-hop, asserts the AS-path carries exactly one AS_SEQ [100, 200] segment, and checks community translation.
- **test_bgp_prefix_states_all_peers_b2b** — empty bgp_peer_names returns state for all configured peers.
- **test_bgp_prefix_states_ipv6_b2b** — same coverage over a dual-stack config.
- **test_bgp_prefix_states_filter_b2b** — 5 prefixes unfiltered, narrowed to 100.1.0.0/24 with an address filter.

### DIFF
--- a/snappi_ixnetwork/device/bgp.py
+++ b/snappi_ixnetwork/device/bgp.py
@@ -337,15 +337,17 @@ class Bgp(Base):
 
     # Column-name aliases tried in order (first hit wins).  IxNetwork has
     # varied slightly across versions; this covers the known variants.
-    _V4_ADDR_COLS    = ("IPv4 Prefix", "IP Address",  "Network Address", "Network", "Network Prefix")
+    _V4_ADDR_COLS    = ("IPv4 Prefix",)
     _V6_ADDR_COLS    = ("IPv6 Address", "IPv6 Prefix", "IP Address", "Network Address", "Network")
-    _NLRI_COLS       = ("NLRI", "Prefix Length", "Prefix")
+    _NLRI_COLS       = ("Prefix Length",)
     _NH_COLS         = ("Next Hop",    "NextHop",   "Next-Hop")
+    _V4_NH_COLS      = ("IPv4 Next Hop",)
+    _V6_NH_COLS      = ("Ipv6 Next Hop",)
     _ORIGIN_COLS     = ("Origin",)
-    _LOCPREF_COLS    = ("Local Pref",  "LocalPref", "Local Preference")
+    _LOCPREF_COLS    = ("Local Preference", "Local Pref", "LocalPref")
     _MED_COLS        = ("MED",         "Multi Exit Discriminator")
     _ASPATH_COLS     = ("AS Path",     "AS-Path",   "AsPath")
-    _COMMUNITY_COLS  = ("Communities", "Community")
+    _COMMUNITY_COLS  = ("Community",   "Communities")
     _PATHID_COLS     = ("Path ID",     "PathId",    "Add Path ID")
 
     def get_bgp_peer_objects(self, peer_names):
@@ -477,13 +479,13 @@ class Bgp(Base):
                     columns = table.Columns
                     if not columns:
                         continue
-                    self.logger.warning(
-                        "_get_learned_table: peer=%r Type=%r "
-                        "Columns=%r RowCount=%d"
-                        % (peer_obj.Name, table.Type,
-                           columns, len(table.Values or []))
-                    )
-                    col_idx = {col: i for i, col in enumerate(columns)}
+                    # self.logger.warning(
+                    #     "_get_learned_table: peer=%r Type=%r "
+                    #     "Columns=%r RowCount=%d"
+                    #     % (peer_obj.Name, table.Type,
+                    #        columns, len(table.Values or []))
+                    # )
+                    col_idx = {col.strip(): i for i, col in enumerate(columns)}
                     for row_vals in (table.Values or []):
                         rows.append({
                             col: row_vals[i]
@@ -495,11 +497,6 @@ class Bgp(Base):
                 "_get_learned_table: Table.find() failed for peer %r: %s"
                 % (peer_obj.Name, e)
             )
-
-        self.logger.warning(
-            "_get_learned_table: peer=%r family=%s total_rows=%d"
-            % (peer_obj.Name, family, len(rows))
-        )
         return rows
 
     # --- static low-level helpers ------------------------------------
@@ -551,13 +548,15 @@ class Bgp(Base):
         token = cell.strip()
         while i < len(token):
             ch = token[i]
-            if ch in ("{", "(", "["):
-                close = {"{"  : "}",
-                         "("  : ")",
-                         "["  : "]"}[ch]
-                seg_type = {"{"  : "as_set",
-                            "("  : "as_confed_seq",
-                            "["  : "as_confed_set"}[ch]
+            if ch in ("{", "(", "[", "<"):
+                close = {"{": "}",
+                         "(": ")",
+                         "[": "]",
+                         "<": ">"}[ch]
+                seg_type = {"{": "as_set",
+                            "(": "as_confed_seq",
+                            "[": "as_confed_set",
+                            "<": "as_seq"}[ch]
                 _flush_seq()
                 end = token.find(close, i + 1)
                 inner = token[i + 1 : end if end != -1 else len(token)]
@@ -567,7 +566,7 @@ class Bgp(Base):
             else:
                 # Collect plain ASNs up to the next group delimiter
                 next_group = len(token)
-                for delim in ("{", "(", "["):
+                for delim in ("{", "(", "[", "<"):
                     pos = token.find(delim, i)
                     if pos != -1 and pos < next_group:
                         next_group = pos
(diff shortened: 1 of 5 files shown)


snappi_ixnetwork/device/bgp.py
```diff
@@ -299,6 +320,548 @@ def _configure_bgpv6_route(self, v6_routes, ixn_bgp):
...
+        """
+        requested = set(peer_names) if peer_names else None
+        ixn_object_names = set(self._ngpf.api.ixn_objects.names)
+        results = []
+
+        topologies = self._ngpf.api._ixnetwork.Topology.find()
+        dg_chain = topologies.DeviceGroup.find().Ethernet.find()
+
+        try:
+            bgpv4_peers = dg_chain.Ipv4.find().BgpIpv4Peer.find()
```
alakendu, line 466: This approach may become slow when the DG is very large. It may also not work correctly if BGP is configured on top of a loopback or chained DG, although I'm not sure whether snappi_ixnetwork supports that scenario.

One alternative could be to use self._ngpf.api.ixn_objects, where all object names are already stored. However, the challenge is that we would need to filter only bgpIpv4Peer objects, and that capability is not currently available.

We can proceed with the current implementation for now, but it would be good to add a note about these limitations. A more robust solution would likely require some restructuring in the future.

snappi_ixnetwork/device/bgp.py
```diff
@@ -299,6 +320,548 @@ def _configure_bgpv6_route(self, v6_routes, ixn_bgp):
...
+        "?": "incomplete",
+        # accept long-form in case a future IxN version emits them
+        "igp": "igp",
+        "egp": "egp",
+        "incomplete": "incomplete",
+    }
+
+    # Column-name aliases tried in order (first hit wins).  IxNetwork has
+    # varied slightly across versions; this covers the known variants.
+    _V4_ADDR_COLS    = ("IPv4 Prefix",)
```
alakendu: I'm not sure why multiple names are being mapped to the same property. As far as I understand, the column display name should remain the same as defined in the LearnedInfo XML.
I'm not entirely sure whether RestPy returns the name property as well. If it does, it would be better to use name instead of displayName, since displayName can sometimes be modified, whereas name typically remains unchanged.

snappi_ixnetwork/device/ngpf.py
```diff
@@ -403,13 +403,60 @@ def get_states(self, request):
...
+            ``{"choice": "bgp_prefixes", "bgp_prefixes": [...]}``, where
+            each list entry is an OTG ``BgpPrefixesState`` dict.
+        """
+        bgp_prefix_req = request.bgp_prefixes
+        peer_names = bgp_prefix_req.bgp_peer_names or []
+        self.logger.debug(
+            "get_bgp_prefix_states peer_names=%s" % peer_names
+        )
+
+        peer_entries = self._bgp.get_bgp_peer_objects(peer_names)
```
alakendu, line 437: I think the current implementation may not be considering the prefix_filters present in the get_state request.

snappi_ixnetwork/device/bgp.py
```diff
@@ -299,6 +320,548 @@ def _configure_bgpv6_route(self, v6_routes, ixn_bgp):
...
+
+        Returns
+        -------
+        list[dict[str, str]]
+            One ``{column_name: value_str}`` dict per prefix row.
+        """
+        # --- Step 1: trigger the learned-info fetch -----------------------
+        # Equivalent to right-click → "Get Learned Info" in the GUI.
+        try:
+            peer_obj.GetAllLearnedInfo()
```
alakendu, line 577: Hopefully, this works since we currently only have IPv4 and IPv6 unicast_prefixes. However, we may need to handle the table index differently if MPLS support is added in the future.

snappi_ixnetwork/device/bgp.py
```diff
@@ -299,6 +320,548 @@ def _configure_bgpv6_route(self, v6_routes, ixn_bgp):
                 self._ngpf.set_device_info(route, ixn_ip_pool)
                 self._configure_route(route, ixn_route)
 
+    # ------------------------------------------------------------------
+    # Learned-info helpers (used by ngpf.get_bgp_prefix_states)
+    # ------------------------------------------------------------------
+
+    # IxNetwork single-char origin → OTG origin string
+    _IXN_ORIGIN_MAP = {
+        "i": "igp",
```
alakendu: I'm not sure how i, e, and ? are expected to be mapped. Hopefully, these characters are not present in the XML metaschema.

tests/bgp/test_bgp_prefix_states.py
```diff
@@ -0,0 +1,345 @@
...
+    assert p.ipv4_address.startswith("100.1.0.")
+    assert p.prefix_length == 24
+    assert p.origin == "egp"
+    assert p.multi_exit_discriminator == 50
+
+    # AS-path must carry exactly one AS_SEQ segment [100, 200]
+    segs = p.as_path.segments
+    assert len(segs) == 1
+    assert segs[0].type == "as_seq"
+    assert segs[0].as_numbers == [100, 200]
```
alakendu, line 364: Please add validation for different AS types, AS numbers, and community values as well. Since a significant amount of parsing logic is involved, it can be difficult to identify issues without validating a wider range of input scenarios.

snappi_ixnetwork/device/bgp.py
```diff
@@ -299,6 +320,548 @@ def _configure_bgpv6_route(self, v6_routes, ixn_bgp):
...
+                except (ValueError, IndexError):
+                    self.logger.warning(
+                        "1. Skipping unrecognised community token: %s" % token
+                    )
+            else:
+                self.logger.warning(
+                    "2. Skipping unrecognised community token: %s" % token
+                )
+        return result
+
```
alakendu, line 925: I do not see any conversion/parsing logic for extended_communities. Is this intentionally omitted, or is it handled elsewhere?

tests/bgp/test_bgp_prefix_states.py
indraniBh: Added UT covering those scenarios.

snappi_ixnetwork/device/bgp.py
indraniBh: extended_communities is not supported yet. It's in todo list.

snappi_ixnetwork/device/bgp.py
indraniBh: There is no name list for columns — columns holds the display names, so keying off the display name is the only option the API offers.
I made a concise list of it now based on live capture.

snappi_ixnetwork/device/bgp.py
indraniBh: Fxed.

snappi_ixnetwork/device/bgp.py
indraniBh: Yes, Documented as limitations.

snappi_ixnetwork/device/bgp.py
indraniBh: Done. Also added TODO in placeholder for mpls for future.

snappi_ixnetwork/device/ngpf.py
indraniBh: Fixed

